PR #1, opened 2018-05-07
Extract Nav component that has a mobile view

dagda1:
Before I go any further, I'd like to know that this is the type of thing you are looking for.

The Sidebar moves to the top at a breakpoint of 600px.

I'll await your feedback before going any futher.

I have some tidying up and styling of the nav in mobile view but lets see if this is what you want first.


example/package.json
```diff
@@ -5,6 +5,7 @@
   "private": true,
   "license": "MIT",
   "dependencies": {
+    "font-awesome": "^4.7.0",
```
bvaughn: nit: Looks like you forgot to update the Yarn lock file.

That being said, I'd like to avoid pulling in a dependency on font-awesome's stylesheets if avoidable. What is this giving us besides an arrow icon? (Can we just make our own?)

bvaughn: As a side note, I kind of prefer the hamburger (3 horizontal line) icon style to the spinning arrow anyway 😄 We could probably just snag this from Google's Material SVGs.

example/src/App.js
```diff
@@ -1,78 +1,132 @@
...
-import FixedSizeListExample from "./routes/examples/FixedSizeList";
-import FixedSizeListWithScrollingIndicatorExample from "./routes/examples/FixedSizeListWithScrollingIndicator";
-import ScrollToItemExample from "./routes/examples/ScrollToItem";
+import DyanmicListExample from './routes/examples/DynamicList';
+import FixedSizeGridApi from './routes/api/FixedSizeGrid';
+import FixedSizeListApi from './routes/api/FixedSizeList';
+import FixedSizeGridExample from './routes/examples/FixedSizeGrid';
+import FixedSizeListExample from './routes/examples/FixedSizeList';
+import FixedSizeListWithScrollingIndicatorExample from './routes/examples/FixedSizeListWithScrollingIndicator';
+import ScrollToItemExample from './routes/examples/ScrollToItem';
```
bvaughn: All of these formatting changes probably weren't intentional?

dagda1:
@bvaughn removed font-awesome and used hamburger.  Will have another play tonight

example/src/App.js
dagda1: no they were not intentional.  I'm using vscode and I thought it would have picked up the .prettierrc.  Where is your setting for double quote strings?

bvaughn:
Sounds great. No hurry.

example/src/App.js
bvaughn: Hm. I think it's actually a problem with my Code settings. 😅 My only concern at the moment is that it will complicate merging. So I should get it sorted out in master I guess.

dagda1:
@bvaughn I would normally use sass so I can be consistent about media queries and use vars for breakpoints, e.g.

```
@mixin mq-xs {
  @include mq($from: gel-bp-xs) {
    @content;
  }
}
```

But sass might be off putting for some people.  I've used styled components but I am not a fan of css-in-js.

Plain css feels weird but maybe for now it is fine.

bvaughn:
Not a fan of using SASS or LESS.

I think at this point, I'm not ready to commit to ejecting from create-react-app so CSS modules is probably the best available option. I might look into trying CSS blocks or just using Glamor again if I decided to eject for other reasons.

I'd say let's not worry about this for now. 😄

dagda1:
@bvaughn how about that for a first move to a more responsive layout.  I've got a pure css hamburger menu now on the left and I've grabbed the available width for the body below the menu in mobile view.

Is there anything else you thing is needed?

dagda1:
Another idea is to have an overlay rather than a drawer that slides out from the side (which I hate) or my implementation of the top sliding down is to have an overlay menu like [this](https://tympanus.net/Development/FullscreenOverlayStyles/)

There is no reason why the sidebar menu should not take over the whole screen in mobile view.

dagda1:
Hi @bvaughn, any feedback?

bvaughn:
Sorry, @dagda1. This is a side project effort for me. My feedback will be a little slow.

I like the new hamburg icon style better. 😄 

There are some changes to `yarn.lock` and `package-lock.json` files in this PR now that I would not expect. Maybe revert them?

> Another idea is to have an overlay rather than a drawer that slides out from the side (which I hate) or my implementation of the top sliding down is to have an overlay menu like this

I think the behavior you linked too looks reasonable. It's similar to how the reactjs.org mobile nav works!

dagda1:
@bvaughn I was not trying to hurry you up :).  I do like that menu.  I'll try that and see what it looks like.

bvaughn:
Thanks again for this PR 😄 I'm going to merge it and play around with it some.

dagda1:
👍

bvaughn:
Pushed to react-virtualized-v10.now.sh after iterating a bit! Thanks again!

dagda1:
@bvaughn looks good.  I think the lesson learned for me is that a sidebar like this in mobile view should just take up the whole screen.  Absolutely no reason not.

bvaughn:
Agreed! Especially as I add more links in the future, I think this will
scale better.

On May 12, 2018 11:38 AM, "Paul" <notifications@github.com> wrote:

@bvaughn <https://github.com/bvaughn> looks good. I think the lesson
learned for me is that a sidebar like this in mobile view should just take
up the whole screen. Absolutely no reason not.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/bvaughn/react-virtualized-v10/pull/1#issuecomment-388574566>,
or mute the thread

PR #2, opened 2018-05-08
basic media queries, to start with.

albinotonnina:
Let's start with some basic media queries.
How would you like to have the SideNav?
Hidden on small devices with a toggler and slide right on click?

image

image


bvaughn:
No worries!

Maybe you two could collaborate on the other one, since it's further along?

PR #3, opened 2018-05-26
Prefer esm over es

TrySound:
`es` is not the best word to identify es modules. It's used by
`material-ui` library to distribute es6+ code. `esm` is a better
alternative which is used by awesome esm loader. In the latest release
rollup started to adrop `esm` word instead of `es`, for now just as an
alias.

`jsnext:main` entry point always was rollup specific and was deprecated
(not officially until we have a real standard) a year ago in favor of
universal `module` field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,16 +5,18 @@
   "author": "bvaughn",
   "license": "MIT",
   "repository": "bvaughn/react-virtualized-v10",
-  "main": "dist/index.js",
-  "module": "dist/index.es.js",
-  "jsnext:main": "dist/index.es.js",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "precommit": "lint-staged",
     "prettier": "prettier --write {example,src}/**/*.{js,jsx,json,css}",
     "linc": "lint-staged",
     "test": "cross-env CI=1 react-scripts test --env=jsdom",
     "test:watch": "react-scripts test --env=jsdom",
-    "build": "rollup -c",
+    "build": "del dist && rollup -c",
     "start": "rollup -c -w",
     "prepare": "yarn run build",
     "predeploy": "cd example && yarn install && yarn run build",
@@ -40,6 +42,7 @@
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-0": "^6.24.1",
     "cross-env": "^5.1.4",
+    "del-cli": "^1.1.0",
     "eslint": "^4.19.1",
     "eslint-config-standard": "^11.0.0",
     "eslint-config-standard-react": "^6.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,16 +5,18 @@
   "author": "bvaughn",
   "license": "MIT",
   "repository": "bvaughn/react-virtualized-v10",
-  "main": "dist/index.js",
-  "module": "dist/index.es.js",
-  "jsnext:main": "dist/index.es.js",
+  "main": "dist/index.cjs.js",
```
bvaughn: This is fine. Seems kind of like a subjective nit but I'm not really opposed to it.

rollup.config.js
```diff
@@ -16,7 +16,7 @@ export default {
     },
     {
       file: pkg.module,
-      format: 'es'
+      format: 'esm'
```
bvaughn: > `es` is not the best word to identify es modules. It's used by
`material-ui` library to distribute es6+ code. `esm` is a better
alternative which is used by awesome esm loader. In the latest release
rollup started to adrop `esm` word instead of `es`, for now just as an
alias.

This seems to be ... somewhat subjective. Even the related Rollup issue (rollup/rollup/issues/1917) says they'll add "esm" as an alias for "es" but they aren't ready to make it _preferred_ or the _default_. (The [Rollup docs also show "es"](https://rollupjs.org/guide/en#configuration-files).)

I don't think we need to update this until/unless it becomes Rollup's recommended format.

package.json
```diff
@@ -5,16 +5,18 @@
   "author": "bvaughn",
   "license": "MIT",
   "repository": "bvaughn/react-virtualized-v10",
-  "main": "dist/index.js",
-  "module": "dist/index.es.js",
-  "jsnext:main": "dist/index.es.js",
```
bvaughn: > `jsnext:main` entry point always was rollup specific and was deprecated
(not officially until we have a real standard) a year ago in favor of
universal `module` field.

What is the harm in specifying both fields, as the repo currently does? Why remove this?

package.json
TrySound: Yeah, it's just a subjective convention: cjs/esm/umd

rollup.config.js
TrySound: It's official api. Guys just trying to make less breaking changes. With a time `es` will be deprecated.

TrySound: Yeah, I just need to update docs.

package.json
TrySound: Because it's useless. `rollup-plugin-node-resolve` by default prefers `module` field. Would be good to not have old stuff in major releases. People look at big projects configurations and follow them.

rollup.config.js
bvaughn: > With a time `es` will be deprecated.

Where is this said?

TrySound: They are open to deprecation but we need a time because option can be used by plugins.
https://github.com/rollup/rollup/pull/2102

TrySound: I think I will manage this process.

bvaughn: I see. Okay then. 👍

PR #4, opened 2018-05-26
Add size snapshot

TrySound:
This plugin shows a few different sizes and provides treeshakability
information (some more stuff is coming). This snapshot also can
demonstrate users how much the project will add to their bundles.

Size snapshot is generated only with build script to not freeze watch
mode.


.size-snapshot.json
```diff
@@ -0,0 +1,21 @@
...
+    "gzipped": 4817
+  },
+  "dist/index.esm.js": {
+    "bundled": 47513,
+    "minified": 20567,
+    "gzipped": 4746,
+    "treeshaked": {
+      "rollup": {
+        "code": 15990,
+        "import_statements": 21
```
bvaughn: What does this information mean? (How do users consume it?)

bvaughn:
Rebase or merge if you don't mind. I realized my Prettier glob pattern wasn't working right. Had meant to fix that earlier but...

.size-snapshot.json
TrySound: It's import statements size. When `code` and `import_statements` are equal then everything is perfect. I probably need to work more on user friendliness. For now I mostly use it just to achieve treeshakability.

TrySound:
Done

.size-snapshot.json
bvaughn: > When `code` and `import_statements` are equal then everything is perfect.

I'm not sure I understand. The two are nowhere near equal. Can you maybe say this in another way?

To be clear, I see the value of tracking snapshot size. React does this. I think it would be nice to even setup a bot that comments on PRs showing how much each PR would impact the size (also like React does).

I'm just not sure I understand this "treeshaked" info.

(I'm not even sure about the term "treeshaked". Is that even a real term? 😁)

TrySound: The idea is to show how big will be the bundle if you import nothing. Practically this mean a perfect bundle without toplevel side effects. I achieve this via bundling `import {} from 'lib'` entry point. The result size is provided for both rollup and webpack.

`code` means the size of treeshaked bundle
`import_statements` means the size of all `import blaBla from 'bla-bla'` statements which are not treeshaaked by default being external.

TrySound: You import List component here.

```js
import { List } from 'react-virtualized';

<List />
```

react-virtualized has independent List and Grid components

List and Grid uses static properties on classes which makes them not treeshakable so use bundle will still have both List and Grid sources.
```js
const Grid = (() => {})()
Grid.defaultProps = {}
const List = (() => {})()
List.defaultProps = {}

<List />
```

If you use `import {} from 'react-virtualized'` you get something similar
```js
const Grid = (() => {})()
Grid.defaultProps = {}
const List = (() => {})()
List.defaultProps = {}
```
This bundle is used to get treeshaked size

If our code is untranspiled user bundle will get this code which is treeshakable
```js
const List = class List {}
List.defaultProps = {}

<List />
```

If we use empty exports trick then both Grid and List are treeshaked out and we get `0` or some different number if exports are presented. import_statements are computed to figure out that final bundle has only them.

TrySound: This plugin allows us to see the problem but for now it allows me to prove every change I do to make the project treeshakable. In the future I plan to provide an information how to fix problems.

bvaughn: > List and Grid uses static properties on classes which makes them not treeshakable so use bundle will still have both List and Grid sources.

Hm. I don't think this is true?

https://rollupjs.org/repl?version=0.59.3&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMiUyRiolMjBUUkVFLVNIQUtJTkclMjAqJTJGJTVDbmltcG9ydCUyMCU3QiU3RCUyMGZyb20lMjAnLiUyRmNvbXBvbmVudHMuanMnJTNCJTIyJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmNvbXBvbmVudHMuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwUmVhY3QlMkMlMjAlN0JDb21wb25lbnQlN0QlMjBmcm9tJTIwJ3JlYWN0JyUzQiU1Q24lNUNuY29uc3QlMjBGb28lMjAlM0QlMjBwcm9wcyUyMCUzRCUzRSUyMCU3QiU3RCUzQiU1Q25Gb28uZGVmYXVsdFByb3BzJTIwJTNEJTIwJTdCJTdEJTNCJTVDbiU1Q25jbGFzcyUyMEJhciUyMGV4dGVuZHMlMjBDb21wb25lbnQlMjAlN0IlN0QlNUNuQmFyLmRlZmF1bHRQcm9wcyUyMCUzRCUyMCU3QiU3RCUzQiU1Q24lNUNuZXhwb3J0JTIwJTdCRm9vJTJDJTIwQmFyJTdEJTNCJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlMkMlMjJhbWQlMjIlM0ElN0IlMjJpZCUyMiUzQSUyMiUyMiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE

bvaughn: Thanks for talking through this by the way. You know more about this than I do. 😄

bvaughn: > This plugin allows us to see the problem but for now it allows me to prove every change I do to make the project treeshakable. In the future I plan to provide an information how to fix problems.

That's fine, I guess. I agree that it's important to be able to spot such problems. I just wanted to better understand the numbers I was seeing and how I should interpret them or act on them.

TrySound: Thanks fpr answers. Nobody asks so much but I don't think they understand what I'm doing. It's hard for me to formulate all things in my mind.

That's a trick. I use rollup/replace/uglify similar to webpack production mode. Uglify is able to remove pure annotated stuff.

bvaughn: No problem 😄

> I use rollup/replace/uglify similar to webpack production mode. Uglify is able to remove pure annotated stuff.

This sounds like something that libraries which _use_ `react-virtualized` would need to do, and not something we'd want to do as part of the build process here. Am I understanding that correctly?

TrySound: Well, you need to use replace plugin to define NODE_ENV which necessary in react apps. And of course you should minify the code in production. Webpack does this with `-p` flag.

bvaughn: Right, right. Sorry if I was being unclear. Let me try asking another way.

Based on the numbers your plug-in shows here, do you see things that this project is doing (v10) that can be improved?

TrySound: Yeah, sure. Any code can be improved. Snapshot just shows current status and prevents later regressions.

The biggest problem of react for treeshakability is static properties because they are rendered outside of class iife. But I have an idea how to fix it.

bvaughn: > Any code can be improved.

Yes, this is super vague though. I was asking _what things_ 😄

> The biggest problem of react for treeshakability is static properties

Still not sure what you mean by this. Did you look at the Rollup REPL I linked to? Static properties alone [don't cause problems](https://rollupjs.org/repl?version=0.59.3&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMiUyRiolMjBUUkVFLVNIQUtJTkclMjAqJTJGJTVDbmltcG9ydCUyMCU3QiU3RCUyMGZyb20lMjAnLiUyRmNvbXBvbmVudHMuanMnJTNCJTIyJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmNvbXBvbmVudHMuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwUmVhY3QlMkMlMjAlN0JDb21wb25lbnQlN0QlMjBmcm9tJTIwJ3JlYWN0JyUzQiU1Q24lNUNuY29uc3QlMjBGb28lMjAlM0QlMjBwcm9wcyUyMCUzRCUzRSUyMCU3QiU3RCUzQiU1Q25Gb28uZGVmYXVsdFByb3BzJTIwJTNEJTIwJTdCJTdEJTNCJTVDbiU1Q25jbGFzcyUyMEJhciUyMGV4dGVuZHMlMjBDb21wb25lbnQlMjAlN0IlN0QlNUNuQmFyLmRlZmF1bHRQcm9wcyUyMCUzRCUyMCU3QiU3RCUzQiU1Q24lNUNuZXhwb3J0JTIwJTdCRm9vJTJDJTIwQmFyJTdEJTNCJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlMkMlMjJhbWQlMjIlM0ElN0IlMjJpZCUyMiUzQSUyMiUyMiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE). Static properties combined with IIFEs [do](https://rollupjs.org/repl?version=0.59.3&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMiUyRiolMjBUUkVFLVNIQUtJTkclMjAqJTJGJTVDbmltcG9ydCUyMCU3QiU3RCUyMGZyb20lMjAnLiUyRmNvbXBvbmVudHMuanMnJTNCJTIyJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmNvbXBvbmVudHMuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwUmVhY3QlMkMlMjAlN0JDb21wb25lbnQlN0QlMjBmcm9tJTIwJ3JlYWN0JyUzQiU1Q24lNUNuY29uc3QlMjBGb28lMjAlM0QlMjAocHJvcHMlMjAlM0QlM0UlMjAlN0IlN0QpKCklM0IlNUNuRm9vLmRlZmF1bHRQcm9wcyUyMCUzRCUyMCU3QiU3RCUzQiU1Q24lNUNuY2xhc3MlMjBCYXIlMjBleHRlbmRzJTIwQ29tcG9uZW50JTIwJTdCJTdEJTVDbkJhci5kZWZhdWx0UHJvcHMlMjAlM0QlMjAlN0IlN0QlM0IlNUNuJTVDbmV4cG9ydCUyMCU3QkZvbyUyQyUyMEJhciU3RCUzQiUyMiU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJjanMlMjIlMkMlMjJuYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==), but that's not inherent to React.

bvaughn: What's the strategy for using the built snapshot JSON in this repo?

Is the file committed to `master` by CI, and compared for branches (but not committed)? Or...?

TrySound: There is `machSnapshot` option which fails if generated snapshot does not match existing one with optional threshold. With CI this forces users to commit new size. Maybe bot is good too but I never worked with them.

bvaughn: I'm a little wary of a programmatic threshold. I like the idea of a bot that comments on a PR with a diff (like React does currently) and then a maintainer can decide if the feature is worth the delta.

TrySound: I didn't threshold too but added because of a couple of requests. The idea is the same but instead of bot snapshot is committed by user. And user can be forced by CI. Threshold is 0 by default so user is always forced to update snapshots.

TrySound:
So what do you decide? I need to track treeshaked size to see improvements.

bvaughn:
I still feel the same way as [before](https://github.com/bvaughn/react-virtualized-v10/pull/4#discussion_r191110654).

I don't see the usefulness of this change without a CI+PR bot strategy.

And I'm not comfortable with a programmatic threshold. I think it's useful info but a maintainer needs to make the call.

I do think it's useful to track over time though.

TrySound:
Take a look at the latest snapshot. This is what I wanted to achieve with visible results.

bvaughn:
I understand the usefulness of such a metric 😄 React uses a similar approach to measure the impact of changes on build size (with a bot that comments on PRs). I'd love to have something like that for react-window.

My concern all along, as I've mentioned, is that I'm not comfortable with a programmatic threshold that fails the build (i.e. the `machSnapshot` option). The information should be highly visible, but a maintainer should make the call about whether to merge or not.

I guess the end-to-end story of how this would work is a bit unclear to me. At the end of our last thread, you said:
> The idea is similar but instead of bot snapshot is committed by user. And user can be forced by CI. Threshold is 0 by default so user is always forced to update snapshots.

I think the ergonomics of relying on / requiring users to submit this file seems less than ideal. It would also cause merge conflicts all over the place any time more than one PR is opened. I prefer the approach React uses where CI runs the task locally and leaves a comment with the delta. Would something like that be possible?

bvaughn:
Ideally, we would also measure the impact of PRs like #25 which strip code from production bundles as well.

TrySound:
Production/development sizes is a good idea.

I already started investigation about bots. And still I would like to make its result visible in one place so users could rely on it choosing library. Do you have ideas where from bot perspective?

TrySound:
Could you submit an issue about dev prod sizes?

bvaughn:
> I already started investigation about bots. And still I would like to make its result visible in one place so users could rely on it choosing library. Do you have ideas where from bot perspective?

Interesting. Sounds like you're talking about a website (similar to bundle phobia) that shows the results for various packages, where as I've been focusing on how a maintainer would use this data to avoid regressions.

> Could you submit an issue about dev prod sizes?

Sure! https://github.com/TrySound/rollup-plugin-size-snapshot/issues/20

TrySound:
Yep, probably a link in readme, or regenerated image could help here.

PR #11, opened 2018-06-12
fix item search

nihgwu:
the `lastMeasuredIndex` could be `-1` after `resetAfterIndex`, then `findNearestItemExponentialSearch` would crash because `getItemMetadata(itemType, props, index, instanceProps)` is undefined


bvaughn:
Thanks

PR #18, opened 2018-07-22
Use babel loose mode

Andarist:
```sh
npx terser -cm --toplevel dist/index.esm.js | gzip -c | wc -c
```

pre - 5828 bytes
post - 5369 bytes

reduction 8% 💰

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,8 @@
 {
   "presets": [
     ["env", {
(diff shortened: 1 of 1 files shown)


bvaughn:
Not sure how I feel about this one. Do you have a sense fo what % of libraries use loose mode and how often they run into spec non-compliance problems?

PR #25, opened 2018-07-23
More granular env checks - micro output size optimization

Andarist:
Micro optimization - from 4.92 KB to 4.89 KB

**NOTE** I suspect this might not pass flow check now (might need to drop some casts to any because of mutating class after declaration), but didn't have time to investigate this, because this repo has no flow setup (relaying on globally installed flow?). Would be good to add flow as dependency and create a task for it.

### DIFF
--- a/src/FixedSizeGrid.js
+++ b/src/FixedSizeGrid.js
@@ -4,7 +4,7 @@ import createGridComponent from './createGridComponent';
 
 import type { Props, ScrollToAlign } from './createGridComponent';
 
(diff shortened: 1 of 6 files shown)


bvaughn:
Not sure how I feel about this change 😄 I think the code is slightly more readable the way it was before, and the gains are super small. Going to think about it.

bvaughn:
I appreciate the work done on this PR ❤️ but I think I'm going to pass on it for now. Let's talk more if you think I made the wrong decision and there's a compelling reason for landing it.

PR #32, opened 2018-08-02
add shouldForceUpdate to resetAfterIndex

nihgwu:
- [x] api
- [x] docs
- [x] tests

🤓

### DIFF
--- a/src/__tests__/VariableSizeGrid.js
+++ b/src/__tests__/VariableSizeGrid.js
@@ -206,19 +206,6 @@ describe('VariableSizeGrid', () => {
   });
 
   describe('resetAfterIndex method', () => {
-    it('should not call forceUpdate if shouldForceUpdate is false', () => {
-      const rendered = ReactTestRenderer.create(
-        <VariableSizeGrid {...defaultProps} />
-      );
-      rendered.getInstance().forceUpdate = jest.fn();
-      rendered.getInstance().resetAfterIndices({
-        columnIndex: 5,
-        rowIndex: 15,
-        shouldForceUpdate: false,
-      });
-      expect(rendered.getInstance().forceUpdate).toHaveBeenCalledTimes(0);
-    });
-
     it('should recalculate the estimated total size', () => {
       const columnWidth = jest.fn(() => 75);
       const rowHeight = jest.fn(() => 35);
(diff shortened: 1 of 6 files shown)


website/src/routes/api/VariableSizeGrid.js
```diff
@@ -96,39 +96,59 @@ const PROPS = [
...
+        <p>
+          <code>VariableSizeGrid</code> caches offsets and measurements for each
+          column index for performance purposes. This method clears that cached
+          data for all columns after (and including) the specified index. It
+          should be called whenever a column's width changes. (Note that this is
+          not a typical occurrance.)
+        </p>
+        <p>
+          You can set <code>shouldForceUpdate</code> to <code>false</code>
+          to prevent the list calling <code>forceUpdate</code> internally.
```
bvaughn: I don't think this is enough of an explanation. Maybe something more like:

> By default the grid will automatically re-render after the index is reset. If you would like to delay this re-render until e.g. a state update has completed in the parent component, specify a value of `false` for the second, optional parameter.

(And similar for the below blocks.)

src/__tests__/VariableSizeGrid.js
```diff
@@ -206,6 +206,19 @@ describe('VariableSizeGrid', () => {
...
+      const rendered = ReactTestRenderer.create(
+        <VariableSizeGrid {...defaultProps} />
+      );
+      rendered.getInstance().forceUpdate = jest.fn();
+      rendered.getInstance().resetAfterIndices({
+        columnIndex: 5,
+        rowIndex: 15,
+        shouldForceUpdate: false,
+      });
+      expect(rendered.getInstance().forceUpdate).toHaveBeenCalledTimes(0);
```
bvaughn: I think the way the "_should recalculate the estimated total size_" test checks for this is a bit more robust. (It checks to see if the `columnWidth` or `rowHeight` getter functions have been called.)

`forceUpdate` is kind of an implementation detail (e.g. maybe the grid re-renders using `setState` instead?)

bvaughn:
Thanks for the PR! Requested some minor changes. 😄

src/__tests__/VariableSizeGrid.js
```diff
@@ -258,6 +245,63 @@ describe('VariableSizeGrid', () => {
...
+          overscanCount={1}
+          columnWidth={index => 40}
+          rowHeight={index => 20}
+        />
+      );
+      // The estimated total height should be (100 + 20 * 1 + 30 * 14)px = 540px.
+      // The estimated total width should be (200 + 40 * 1 + 30 * 4)px = 360px.
+      expect(scrollContainer.props.style.height).toEqual(540);
+      expect(scrollContainer.props.style.width).toEqual(360);
+    });
```
bvaughn: This is a much better test! Thank you!

bvaughn:
Published as 1.1.0

PR #40, opened 2018-08-11
Distribute flow types

TrySound:
This simple trick works similary to bundling. One file reexports all
sources with its types.

```js
// @flow

export * from '../src';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
   "files": [
-    "dist"
+    "dist",
+    "src/*.js"
   ],
   "scripts": {
     "flow": "flow check src && flow check website",
@@ -45,7 +46,8 @@
     "lint": "eslint '**/*.js'",
     "test": "cross-env CI=1 react-scripts test --env=jsdom",
     "test:watch": "react-scripts test --env=jsdom",
-    "build": "del dist && rollup -c",
+    "build:flow": "echo \"// @flow\n\nexport * from '../src';\" > dist/index.cjs.js.flow",
+    "build": "del dist && rollup -c && yarn build:flow",
     "start": "rollup -c -w",
     "prepare": "yarn run build",
     "website:build": "cd website && yarn build",


bvaughn:
I don't want to ship the entire source in the NPM bundle.

bvaughn:
I appreciate the suggestion though. A [similar suggestion on react-flame-graph](https://github.com/bvaughn/react-flame-graph/pull/7) prompted my comment earlier on Twitter. I don't find shipping the entire source to NPM a very satisfactory solution.

TrySound:
This is the easiest solution. flow-copy-source is the same publishing sources. NPM module bloat happens when authors publish tests, fixtures, dozens of configs. Sources of this project is just 7 not too big files. There won't be bloat.

Any generation doesn't work in flow well. flow-copy-source makes flow more verbose when types are out of sync with sources and previously built types have an error. So this is just the least hacky solution. I use it in a lot of projects and it just works without hacks and broken packages.

https://unpkg.com/react-beautiful-dnd@9.0.0/dist/
https://unpkg.com/react-powerplug@1.0.0-rc.1/dist/
https://unpkg.com/react-select@2.0.0/lib/
https://unpkg.com/jss@9.8.7/lib/
https://unpkg.com/react-head@3.0.0-0/dist/

TrySound:
@bvaughn Is it worth to bikeshed about distribution and dependency bloat when this library is quite small. Sources inside node_modules won't make anybody's node_modules Bigger.

Can we go this way until we will find a better solution?

bvaughn:
> Sources inside node_modules won't make anybody's node_modules Bigger.

Isn't that exactly what putting sources in `node_modules` does?

I dunno. I still really dislike this, and I'm upset with Flow for not providing a better solution for such an obvious use case.

TrySound:
I meant sources of this library specifically. They are quite small.

bvaughn:
Makes sense :smile:

bvaughn:
Okay. As much as I dislike having to do this, I think it's the only viable option at least for the short term and I appreciate how annoying it must be to _not_ have Flow coverage for this lib, so let's do it. Thanks for your patience.

TrySound:
I'm also thinking about inferring itemData inside item. Currently we have `any` there.

bvaughn:
I dig it.

bvaughn:
Flow types have been included with the [1.2.0 release](https://github.com/bvaughn/react-window/blob/master/CHANGELOG.md#120)

KevinGrandon:
🎉 I haven't used this library yet, but just wanted to say thanks for landing flow typings! We've been following a similar approach for all of our [fusion](https://unpkg.com/fusion-core@1.5.2/dist/index.js.flow) packages and it's worked out quite well for us.

bvaughn:
That's nice to hear @KevinGrandon. Thanks!

PR #48, opened 2018-08-29
Upgrade to babel 7

TrySound:
In this diff I moved build to babel 7, removed unused packages and made
babel helpers importable from runtime. This allows to reuse these
helpers across many packages and save a few kilobytes in user bundle.


bvaughn:
Backwards compatible change otherwise?

This looks great 😁

TrySound:
Sure. Just one dependency added.

PR #52, opened 2018-09-06
Add TypeScript definitions

martynaskadisa:
First of all, thank you for this amazing library!

Since this library is written in Flow and I mainly use TypeScript, I've decided to contribute type definitions.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,351 @@
+declare module 'react-window' {
+    import { Component, ComponentType, CSSProperties, Ref, Key } from 'react';
+
+    type Direction = 'vertical' | 'horizontal';
+    type ScrollDirection = 'forward' | 'backward';
+    type Align = 'auto' | 'center' | 'end' | 'start';
+
+    interface ListChildComponentProps {
+        index: number;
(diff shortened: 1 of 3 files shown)


bvaughn:
Hi @martynaskadisa! Thanks for taking the time to create these 😄

I'm not a TypeScript user myself, so I'm not sure about this– but wouldn't it be a bit more "standard" to submit these to the [Definitely Typed repo](https://github.com/DefinitelyTyped/DefinitelyTyped)? That way other people could help maintain them.

I'm pretty reluctant to merge types here, because then I'm responsible to keep them in sync for future releases– and I'm likely to forget to do this or to make a syntax error, since I'm not a TypeScript user.

bvaughn:
Thanks very much 😄

martynaskadisa:
I've opened a [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28824) in DefinitelyTyped repo. Linking it here as a reference in case anyone else stumbles upon this pull request.

bvaughn:
Thanks again! 😄

PR #77, opened 2018-10-25
Add test and fix for flowtype for children

### LINKED ISSUE #76: Flow Type issue with PureComponent
Hi,

I have an issue with the flow types provided by react-window when I provide a class as a child ( like in the memoized example):

[flow] Cannot create `FixedSizeGrid` element because a callable signature is missing in statics of `ListItem` [1] but exists in `RenderComponent` [2] in property `children`. (References: [1] [2])

Also thanks for the great library 👍

nicholas-l:
I think this fixes it but I'm not sure.

Fixes #76

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -18,7 +18,9 @@ type RenderComponentProps<T> = {|
   rowIndex: number,
   style: Object,
 |};
-export type RenderComponent<T> = (props: RenderComponentProps<T>) => React$Node;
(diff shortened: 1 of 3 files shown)


bvaughn:
Hm. I _think_ this change looks like a good one to make, but the Flow failure still happens for me (in this branch, as well as in a CRA app).

We should make a similar change in `createGridComponent` too.

bvaughn:
This fix has been published in v1.2.2!

Issue #6, opened 2018-05-30
Support just-in-time measured content


bvaughn:
In order to avoid adding cost to existing list and grid components, create a new variant (e.g. `DynamicSizeList` and `DynamicSizeGrid`). This variant should automatically measure its content during the commit phase.

### MVP

The initial implementation of this could work similarly to how `CellMeasurer` works in react-virtualized:
1. Content is measured only if no current measurements exists.
2. Measurements need to be reset externally (imperatively) if something changes.
3. Cells at a given index can only be positioned after all cells before that index have been measured.

### Goal

This component could perform better if we removed the third constraint above, allowing random access (by either item index or scroll offset) without measuring the preceding items. This would make react-window much more performant for use cases like chat applications.

This would also unlock the ability to use a [ResizeObserver](https://rawgit.com/WICG/ResizeObserver/master/index.html) (via [react-measure](https://github.com/souporserious/react-measure)) to automatically detect item sizing and remove the position and measurements cache entirely. This would remove the need for imperatively resetting cached measurements and dramatically improve the API.

In order for the above to be possible, the dynamic list/grid components would need to use a dramatically different approach for mapping offset to index and vice versa. ([This comment about "scroll anchoring" in react-virtualized](https://github.com/bvaughn/react-virtualized/issues/610#issuecomment-324890558) has some nice visuals.) Essentially, we would need to do something like this:

* Estimate total size based on the number of items multiplied by a `estimatedItemSize` prop. (This estimated size won't need to be adjusted, since the mapping described below doesn't is fuzzy.) 

* When scroll position changes, compare the new offset to the previous offset. **If the delta is greater than some [to be determined] threshold**, set the new offset as the "scroll anchor". Map the offset to an estimated index (e.g. divide the offset by total estimated scrollable size and multiply that by the number of items in the collection). Store this mapped index as the "anchor index". For example, if the list described by the image below had 250 items, the "anchor index" would be 132.

<img width="328" alt="screen shot 2018-06-10 at 11 58 38 am" src="https://user-images.githubusercontent.com/29597/41205298-ae8c10a4-6ca5-11e8-8630-ad94913d7112.png">

* When scroll position changes, **if the delta is less than the threshold**, choose which new items to render relative to the anchor index. Position these items relative to the previously positioned items. Continuing with the example above, if the list was scrolled by a small amount (200px) then 200px worth of additional rows would need to be appended below the previously positioned items:

[...]

PR #102, opened 2018-12-07
Just-in-time measured content [work in progress]

bvaughn:
Initial implementation of a list component supporting just-in-time measured content (relates to issue #6).

### Checklist
- [x] Gracefully degrade when `ResizeObserver` API is not available.
- [x] Fallback to using `findDOMNode` when no DOM element ref is available, but warn about this.
- [ ] Ignore resize events when hidden (see [this comment](https://github.com/bvaughn/react-window/issues/6#issuecomment-435372932))
- [ ] Use RAF to batch multiple updates (for MO polyfill, which is async?)
- [ ] RTL support
- [ ] Negative offsets check to fix Safari's elastic scroll issue (#251)
- [ ] Smoke test the following browsers/platforms:
  - [ ] Desktop
    - [x] Chrome
    - [x] Edge
    - [x] Firefox (momentum scrolling caveat for Firefox <= 64, fe88aeb)
    - [ ] Internet Explorer
    - [x] Safari
  - [ ] Mobile
    - [x] Android
    - [ ] iOS (momentum scrolling jumps around)


src/ItemMeasurer.js
```diff
@@ -0,0 +1,107 @@
...
+  // eslint-disable-next-line no-use-before-define
+  (entries: Entries, observer: ResizeObserver): void,
+};
+
+declare class ResizeObserver {
+  constructor(ResizeObserverCallback): ResizeObserver;
+  observe(target: HTMLElement): void;
+  unobserve(target: HTMLElement): void;
+  disconnect(): void;
+}
```
TrySound: Actually flow already has ResizeObserver types :)
https://github.com/facebook/flow/blob/master/lib/bom.js#L446-L456

src/ItemMeasurer.js
```diff
@@ -0,0 +1,107 @@
+// @flow
+
+import { Component } from 'react';
+import { findDOMNode } from 'react-dom';
```
TrySound: Does this mean dynamic measurements won't work with concurrent mode?

src/ItemMeasurer.js
```diff
@@ -0,0 +1,107 @@
...
+    this._node = node;
+
+    // Force sync measure for the initial mount.
+    // This is necessary to support the DynamicSizeList layout logic.
+    this._measureItem(true);
+
+    if (typeof ResizeObserver !== 'undefined') {
+      // Watch for resizes due to changed content,
+      // Or changes in the size of the parent container.
+      this._resizeObserver = new ResizeObserver(this._onResize);
```
TrySound: I use ponyfill and would like to pass this class via props. Are you good with this?

src/ItemMeasurer.js
bvaughn: Not the version I'm using (0.80). [`ResizeObserver` defs were added two months ago](https://github.com/facebook/flow/commit/599e26be97f96744b531d416bde47224e70ae573) so I think that means 0.82 or 0.83. I'll update at some point soon.

src/ItemMeasurer.js
bvaughn: If `findDOMNode` goes away in the future, I will have to switch to a ref-based approach that uses `cloneElement()` but I may leave that for future Brian to deal with.

bvaughn: Meh...maybe I'll add that now and just fallback to `findDOMNode` and log a DEV warning if people don't forward the ref properly

TrySound: Good plan

natew:
Seems like two last bugs right, iOS and resize? And then merge with master? I could help bring this up to date with master, and look at resize.

bvaughn:
@natew If you could rebase on master and fix merge conflicts, that'd be a great help. I've been meaning to find some time to finish this off, but I've been busy lately.

piecyk:
@bvaughn merged master on my fork https://github.com/piecyk/react-window/pull/1

bvaughn:
@piecyk Cool.
Is that the only change in your fork?
Gimme a PR into this branch?

piecyk:
@bvaughn yep only that, there are two commits as first master merge was done some time ago... always testing stuff on other branches 😂Sending PR 🚀

bvaughn:
Thanks!

natew:
@piecyk thank you!

@bvaughn noticed the version is bumped to 1.8.0-alpha but not published, could you publish when you get a chance?

edgars-sirokovs:
> @piecyk thank you!
> 
> @bvaughn noticed the version is bumped to 1.8.0-alpha but not published, could you publish when you get a chance?

Hi! Any chance this will be published in the near future?

zmitry:
Any updates on this ?

Hellycat:
May you inform about approximate date of merging this PR into master branch?

john-osullivan:
In case anybody wants to start using this WIP before @bvaughn feels comfortable publishing to NPM, remember you can directly install this branch's version directly to your project via `npm i bvaugh/react-window:issues/6`.

I was just in the middle of writing up an involved solution for all the refs management, very glad to see the built-in solution is so close to primetime!  You the bomb

john-osullivan:
In case anybody wants to start using this WIP before @bvaughn feels comfortable publishing to NPM, heads up that you can directly install this branch's version directly to your project via `npm i bvaughn/react-window#issues/6`.  Make sure you've also got `yarn` installed, it'll error out otherwise.

I was just in the middle of writing up an involved solution for all the refs management, very glad to see the built-in solution is so close to primetime! You the bomb

natew:
I've found a behavior we may want to avoid.

Basically, I have a virtual list inside a div that will be set to `style={{ display: none }}` when hidden. This triggers the ResizeObserver to update, which causes many extra renders, as well as visual flickering because react-window thinks the list height has gone to 0.

We could make ItemMeasurer customizable as a prop, so I could customize the resize observing logic.

Bessonov:
@natew what is the goal to set `display: none`? For me, it's expected and valid behavior.

natew:
To prevent resizing from triggering paints within the entire parent container that has that list. 

I don’t think you can claim it’s expected given you don’t have the use case. 

Re rendering on display none doesn’t make sense in this case. Further, if you display it again you see a flicker where the list items are all overlapped at top = 0 while it measures and then shows again. So it’s one bug and one extra render. 

One way to think of it is does it behave as similarly to a non-virtual list as possible. A non virtual list wouldn’t re render on display changes. 

I submitted a PR that let’s you override CellMeasurer. It’s not perfect, the bug still will happen without using a custom one, but it works for this case and maybe others.

bvaughn:
The built-in measuring solution needs to handle hidden/display:none as per the TODO item in the PR description:
> Ignore resize events when hidden

I don't think requiring an overridden measurer is an approach I want to take. I'd rather this just work out of the box. I have a rough idea for how we might tackle it.

natew:
Ah missed the todo, my bad. Totally agree that’s a better solution I was
thinking it was intended.

brunolemos:
Tried using `DynamicSizeList` on my project but haven't succeeded. `VariableSizeList` works.

It seems all items are being positioned on top of each other. 
Has anyone faced this or know the possible cause?

image

image

image

The code is here: https://github.com/devhubapp/devhub/pull/152

john-osullivan:
@natew I just ran into an issue like yours because my list lives in an `@reach/tabs` Tab and the implementation sets non-visible tabs to `display:none`.  My list is pretty fat (1k min), so I was getting a full React crash when it was hidden (`Uncaught Invariant Violation: Maximum update depth exceeded.`).  

It's not an ideal solution, but given that we're writing "in-the-meantime" hacks anyways, I found that using tab visibility to switch from a `DynamicSizeList` to a `FixedSizeList` is performant enough and gets around this temporary issue.  You do lose your scroll position as was mentioned in the comment linked on the todo, though, so definitely not a long-term answer.

I know this is just a side project for you @bvaughn ; if you can explain your rough thinking on how you'd like to approach this, I'd be happy to lend a hand with writing up the final solution 👍

john-osullivan:
@brunolemos The maintainer has said he'd prefer to not have questions re usage in here, keeping the discussion focused on implementation.  That said, if you scroll up and check out the sample code I linked, you'll see that my row component uses `React.forwardRef()` to pass a ref onto each row item.  If your rows don't have refs, then `react-window` can't measure them.  Try adding those refs and you'll probably be good to go.

brunolemos:
@john-osullivan thanks, deleted the comment and will delete this one as well. Will try adding refs!

renzosal:
Hi! any chance 1.9.0-alpha.1 could be published?

brunolemos:
@bvaughn found a bug in this branch, if my items are exactly height `50` (the default estimated size) they will all render on top of each other:

image

If they have any different height (e.g. `49` or `51`) it works:

image

src/ItemMeasurer.js
brunolemos: This conversation was not marked as resolved, is react-window DynamicSizeList compatible with concurrent mode? I'm about to enable it.

john-osullivan:
Hallo to fellow react-window users eager for this feature to be released.  I ran into an issue with my CI/CD environment not properly building packages installed from git, so my prior workaround doesn't work for production deployment.  In order to solve it, I forked and published the contents of this branch @ bc9192b as a scoped package:

`npm i @john-osullivan/react-window-dynamic-fork`

In order to prevent myself from forgetting to update when the feature is released, I added a file which runs at buildtime and checks the most recent published version of `react-window`.  If the version satisfies `>=1.9.0`, it throws an error and exits the process telling you to uninstall my fork and put the proper package back in:

```javascript
const npmApi = require('npm-api');
const semver = require('semver');
const process = require('process');
const npm = new npmApi();
const reactWindow = npm.repo('react-window');

reactWindow.package().then((pkgJson) => {
  let version = pkgJson.version;
  if (semver.satisfies(semver.coerce(version), '>=1.9.0')) {
    throw new Error(`react-window ${version} has been released, please uninstall this fork and reinstall react-window.`);
  } else {
    console.log(`\nMost recent react-window version is ${version}, DynamicList not yet available on npm.`);
    console.log('This fork package will inform you when react-window @ 1.9.0 is available. \n')
  }
}).catch((err) => {
  console.log(`\n${err}\n`);
  process.exit(1)
});
```

Obviously use at your own risk, as this PR isn't complete.  Happy hacking! @renzosal @tony-scio @patryksuchowierski @Hellycat @jazevedo620 @satyadeepk

PR #120, opened 2019-01-14
Recover containerStyle feature from react-virtualized

mudafar:
Add `innerStyle` prop  to extend inner element style, just like the outer element.

```javascript
        createElement(innerElementType || innerTagName || 'div', {
          children: items,
          ref: innerRef,
          style: {
            height: direction === 'horizontal' ? '100%' : estimatedTotalSize,
            pointerEvents: isScrolling ? 'none' : '',
            width: direction === 'horizontal' ? estimatedTotalSize : '100%',
            ...innerStyle, //the new prop.
          },
```

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -54,6 +54,7 @@ export type Props<T> = {|
   initialScrollTop?: number,
   innerRef?: any,
   innerElementType?: React$ElementType,
+  innerStyle?: Object,
   innerTagName?: string, // deprecated
   itemData: T,
   itemKey?: (params: {|
@@ -288,6 +289,7 @@ export default function createGridComponent({
         height,
         innerRef,
         innerElementType,
+        innerStyle,
         innerTagName,
         itemData,
         itemKey = defaultItemKey,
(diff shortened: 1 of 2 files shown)


bvaughn:
Hey @mudafar 👋 Thanks for submitting a PR!

This doesn't look like a change I want to land though. (I don't want to go down the road of adding more and more inner* and outer* props to this library. It adds bloat (to the library and docs) and makes project maintenance more difficult.)

You can use the recent `innerElementType` for edge case things like this. 😄 [For example](https://codesandbox.io/s/km5v387zko):

```js
const InnerContainer = ({ style, ...rest }) => (
  <div
    style={{
      ...style,
      backgroundColor: "red"
    }}
    {...rest}
  />
);

const Example = () => (
  <List
    className="List"
    height={150}
    innerElementType={InnerContainer}
    itemCount={1000}
    itemSize={35}
    width={300}
  >
    {Row}
  </List>
);
```

mudafar:
Hi @bvaughn,

Yes you're right and thanks for the example!

PR #138, opened 2019-02-18
Handle case when number of element reduces in list

iamsolankiamit:


bvaughn:
Can you include a repro that fails before this change is made?

iamsolankiamit:
here, https://codesandbox.io/s/64p6x0pkn3

bvaughn:
I see. Thanks for sharing.

I agree with you that this looks like a bug. I'd like to add a unit test for it to prevent a regression. I'm pretty tired tonight. I'll take another look in the morning.

src/VariableSizeList.js
```diff
@@ -147,6 +147,7 @@ const getEstimatedTotalSize = (
   let totalSizeOfMeasuredItems = 0;
 
   if (lastMeasuredIndex >= 0) {
+    if (lastMeasuredIndex >= itemMetadataMap.length) lastMeasuredIndex = itemMetadataMap.length - 1;
```
bvaughn: [`itemMetadataMap` is an `Object`](https://github.com/bvaughn/react-window/blob/a22f3e983dd3d15da1e51230189c928ad93f9bbb/src/VariableSizeList.js#L252) so it doesn't have a `length` property.

bvaughn: I'll just make this adjustment myself and also add the fix to `VariableSizeGrid`. Thanks for pointing this out.

bvaughn:
This fix has been released with version [1.5.2](https://github.com/bvaughn/react-window/blob/master/CHANGELOG.md#152)

PR #150, opened 2019-03-01
Add RTL support for List component

davidgarsan:


davidgarsan:
This changes provide support for RTL direction in List component.
Still pending to add the similar logic to Grid component.

bvaughn:
Thank you for taking the time to open this PR!

Several things will need to happen before I can merge and release this feature though:
- [ ] The `direction` prop will need to be added to grid component as well.
- [ ] The prop will need to be added to the [`FixedSizeList`](https://react-window.now.sh/#/api/FixedSizeList) and [`FixedSizeGrid`](https://react-window.now.sh/#/api/FixedSizeGrid) documentation pages.
- [ ] At least one simple unit tests will need to be added for both `FixedSizeList` and `FixedSizeGrid`.
- [ ] The "_Fixed Size List RTL_" docs page should be renamed to something more generic like "RTL layout" and should have an example of a horizontal list and a grid (similar to the ["Scrolling to an item"](https://react-window.now.sh/#/examples/list/scroll-to-item) page).

src/createListComponent.js
```diff
@@ -457,11 +458,15 @@ export default function createListComponent({
           return null;
         }
 
+        const isRtl = this.props.style && this.props.style.direction === 'rtl';
```
bvaughn: I'm not sure about parsing the style in this way. Seems like RTL pages will likely have a `direction: rtl` on the level of the body rather than the list. Normally I'd say an explicit `direction` prop would be better, but that's already used in the list to differentiate between "horizontal" and "vertical". Hmm...

bvaughn: In CSS this is called `direction` but that's already in use for lists. Ideally I would have maybe used "layout" or something but that ship has sailed, and I'm not really willing to do a major release just to rename this property.

There's a related HTML attribute called `dir` but that doesn't really follow the prop naming convention.

I briefly considered maybe using `direction` anyway and just expanding it to be "vertical", "horizontal", or "horizontal-rtl" but that feels awkward.

I am currently leaning toward an `isRTL` boolean attribute, even though...that's a bit weird too.

bvaughn: Maybe `direction` can accept multiple values, e.g. "vertical ltr" (default), "horizontal ltr", "horizontal rtl", etc.

bvaughn: Or maybe I should just add a new `layout` prop and deprecate `direction` (with a DEV warning) for now.

src/createListComponent.js
```diff
@@ -457,11 +458,15 @@ export default function createListComponent({
...
+
         return {
           isScrolling: true,
           scrollDirection:
             prevState.scrollOffset < scrollLeft ? 'forward' : 'backward',
-          scrollOffset: scrollLeft,
+          scrollOffset: isRtl
+            ? scrollWidth - clientWidth - scrollLeft
+            : scrollLeft,
```
bvaughn: I don't know if this is right. Why should we invert the element's `scrollLeft`? That seems...wrong. Am I thinking about this wrong?

bvaughn: Hm... yeah, this seems incorrect (and unnecessary) to me.

bvaughn: Sorry. I think I was mistaken.

bvaughn:
Let's continue the discussion over on #156

src/createListComponent.js
davidgarsan: In my use case I saw the usage of direction (CSS) in an isolated fashion (component biased). But in a real scenario is true that this can be affected by parent styles. Definitely the last approach seems the way to go.

PR #155, opened 2019-03-04
Upgrade memoize-one

jgoz:
Upgrades `memoize-one` to the latest version `^5.0.0`.

Breaking changes between 3.1.1 and 5.0.0 do not affect `react-window`.

- 4.0 changed the bundled output, which was only technically a breaking change
- [5.0 changed the API for the optional comparer function](https://github.com/alexreardon/memoize-one/releases/tag/v5.0.0), which is not used in `react-window`

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "memoize-one": "^3.1.1"
+    "memoize-one": "^5.0.0"
   },
   "peerDependencies": {
     "react": "^15.0.0 || ^16.0.0",
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "memoize-one": "^5.0.0"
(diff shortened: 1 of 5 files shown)


bvaughn:
What benefits does this upgrade bring?

Bumping dependencies by a major version seems like a breaking change.

jgoz:
- [4.0.1](https://github.com/alexreardon/memoize-one/releases/tag/v4.0.1) defines the behavior when the memoized function throws (was previously undefined)
- [4.0.2](https://github.com/alexreardon/memoize-one/releases/tag/v4.0.2) improves performance

I suppose this would be a breaking change for those who are using `--flat` for their dependencies and also depend on `memoize-one@3.x` directly.

Could the dependency in `react-window` be changed to a range? E.g., `>=3.1.1 <6`? This project is compatible with all of those versions since it doesn't use the optional equality function.

bvaughn:
> Could the dependency in react-window be changed to a range? E.g., >=3.1.1 <6? This project is compatible with all of those versions since it doesn't use the optional equality function.

Sure. I guess that would be okay.

jgoz:
Thanks, done.

PR #156, opened 2019-03-04
RTL support

bvaughn:
Builds on top of PR #150 and adds several [additional required changes](https://github.com/bvaughn/react-window/pull/150#issuecomment-469063583).


bvaughn:
Let's do it.

PR #162, opened 2019-03-06
[WIP] Fix RTL scrolling issues in some browsers

davidgarsan:
Attempt to fix some RTL scrolling issues in some browsers due to inconsistent RTL support.
Tested in Chrome, Firefox and Safari latest versions.
The fix is ugly as hell but needs to tackle at least tree different cases:
- Non standard Chrome support for RTL overflow metrics.
- Standard support from Firefox.
- Buggy behaviour on Safari bouncing when the scroll reaches the right limit in RTL layout.

#### Pending tasks:
- Testing in Edge.
- Fix scrollToItem and scrollTo methods to work with RTL direction.
- Apply similar fix for Grid component.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ dist
 .env.test.local
 .env.production.local
 .watchmanconfig
+.idea
 
 package-lock.json
 npm-debug.log*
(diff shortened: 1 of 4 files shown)


bvaughn:
Thanks for this PR!

I think I've found an approach that handles the underlying offset differences and also fixes the imperative scroll-to methods though in PR #251

Going to close this!

PR #167, opened 2019-03-08
Avoid crash if WeakSet is not available

jgoz:
`WeakSet` is not available in IE11, so using `react-window` in development mode causes a runtime error.

This simply falls back to `Set` if `WeakSet` is not available. This could have memory implications since `Set` will hold onto the references indefinitely, but that only affects the one browser and only in development mode. It's also a much smaller impact on bundle size vs., say, a polyfill.

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -4,7 +4,6 @@ import memoizeOne from 'memoize-one';
 import { createElement, PureComponent } from 'react';
 import { cancelTimeout, requestTimeout } from './timer';
 import { getScrollbarSize } from './domHelpers';
-import WeakSet from './weakset';
 
 import type { TimeoutID } from './timer';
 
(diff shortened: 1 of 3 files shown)


bvaughn:
I know this is a small change in terms of LOC, but I'm having a strong initial negative reaction to it. I think because I don't like the idea of leaking memory (even in DEV mode) when there's a reasonable workaround, e.g.:
```js
if (process.env.NODE_ENV !== 'production') {
  if (typeof window.WeakMap === 'undefined') {
    window.WeakMap = require('es6-weak-map/polyfill');
  }
}
```

jgoz:
Yeah, I definitely see your point. What about simply not initializing the sets if `WeakMap` isn't available? IE11 wouldn't get the warnings in dev mode, but at least it wouldn't crash.

bvaughn:
I like this approach _much_ better. Thanks

PR #174, opened 2019-03-12
Make rowIndex and columnIndex optional

### LINKED ISSUE #173: Grid: Scroll to row/column only
For `FixedSizeGrid` and `VariableSizeGrid`, is there a way to only scroll to a particular row OR column index?

I.e., scroll to row 50 while maintaining the current horizontal scroll position.

This kind-of works with `FizedSizeGrid` by using `scrollToItem({ rowIndex: 50, columnIndex: -1 })`, but I'm thinking this just happens to work and is not valid. Trying the same with `VariableSizeGrid` causes an exception.

Specifying `columnIndex: 0` works if the first column does not overflow the outer container, but if it is wider than the container, it will alternate between scrolling to the end of that column and the beginning of the column with `align: "auto"`.

The ideal scenario for my use case would be making `rowIndex` and `columnIndex` optional. If either one is not specified, then `scrollLeft` or `scrollTop` would not be changed from the previous value.

If you would accept a PR for this suggestion I'd be happy to open it.

jgoz:
Closes #173.

Update `scrollToItem` to make `rowIndex` and `columnIndex` optional. If unspecified (or invalid, e.g., negative value), the current value for `scrollTop` or `scrollLeft` will be used.

- Updated tests
- Updated documentation

### DIFF
--- a/src/__tests__/FixedSizeGrid.js
+++ b/src/__tests__/FixedSizeGrid.js
@@ -475,6 +475,10 @@ describe('FixedSizeGrid', () => {
       rendered
         .getInstance()
         .scrollToItem({ columnIndex: 2, rowIndex: 2, align: 'auto' });
+      // Scroll down to row 10, without changing scrollLeft
+      rendered.getInstance().scrollToItem({ rowIndex: 10, align: 'auto' });
+      // Scroll left to column 0, without changing scrollTop
+      rendered.getInstance().scrollToItem({ columnIndex: 0, align: 'auto' });
       expect(onItemsRendered.mock.calls).toMatchSnapshot();
     });
 
@@ -497,6 +501,10 @@ describe('FixedSizeGrid', () => {
       rendered
         .getInstance()
         .scrollToItem({ columnIndex: 99, rowIndex: 99, align: 'start' });
+      // Scroll up to row 10, without changing scrollLeft
+      rendered.getInstance().scrollToItem({ rowIndex: 10, align: 'start' });
+      // Scroll left to column 0, without changing scrollTop
+      rendered.getInstance().scrollToItem({ columnIndex: 0, align: 'start' });
       expect(onItemsRendered.mock.calls).toMatchSnapshot();
     });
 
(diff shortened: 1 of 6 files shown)


src/createGridComponent.js
```diff
@@ -272,22 +272,28 @@ export default function createGridComponent({
...
-        scrollTop: getOffsetForRowAndAlignment(
-          this.props,
-          rowIndex,
-          align,
-          scrollTop,
-          this._instanceProps,
-          horizontalScrollbarSize
-        ),
+        scrollLeft:
+          columnIndex !== undefined && columnIndex >= 0
```
bvaughn: Why is the `>= 0` needed?

jgoz: Well, it's not needed for this feature, but negative indexes will not produce valid results, especially with `VariableSizeGrid`. I can remove the check if you prefer to leave the behaviour undefined.

jgoz: @bvaughn Specifically, for `VariableSizeGrid`, [`itemMetadata`](https://github.com/bvaughn/react-window/blob/master/src/VariableSizeGrid.js#L240) will be `undefined`, which will throw an error [when calculating `maxOffset`](https://github.com/bvaughn/react-window/blob/master/src/VariableSizeGrid.js#L251).

jgoz:
@bvaughn Removed the `>= 0` as it wasn't directly relevant to this enhancement.

src/__tests__/FixedSizeGrid.js
```diff
@@ -475,6 +475,10 @@ describe('FixedSizeGrid', () => {
       rendered
         .getInstance()
         .scrollToItem({ columnIndex: 2, rowIndex: 2, align: 'auto' });
+      // Scroll down to row 10, without changing scrollLeft
+      rendered.getInstance().scrollToItem({ rowIndex: 10, align: 'auto' });
+      // Scroll left to column 0, without changing scrollTop
+      rendered.getInstance().scrollToItem({ columnIndex: 0, align: 'auto' });
```
bvaughn: You aren't testing anything between these `scrollToItem` calls. That seems not great.

bvaughn: Whoops. I misremembered how my own tests were written 😆 you're expecting on the mock calls below. 👍

bvaughn:
Just released with v1.7.0

PR #180, opened 2019-03-16
fix for codemirror loader test regexp not catering for windows paths

glromeo:
I hope this little fix helps others like me trying to build on windows.

### DIFF
--- a/website/config/webpack.config.dev.js
+++ b/website/config/webpack.config.dev.js
@@ -181,7 +181,7 @@ module.exports = {
 
           // Custom CodeMirror loader for syntax highlighting
           {
-            test: /code\/[^/]+$/,
+            test: /code[/\\\\]/,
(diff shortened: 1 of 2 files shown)


website/config/webpack.config.dev.js
```diff
@@ -181,7 +181,7 @@ module.exports = {
 
           // Custom CodeMirror loader for syntax highlighting
           {
-            test: /code\/[^/]+$/,
+            test: /code[/\\\\]/,
```
bvaughn: This is not an equivalent regex. One will match e.g. "code/FixedSizeGridChildrenClass.js" and the other will match e.g. "code/"

bvaughn: Hm. I guess there are other places in the file using the `[/\\\\]` expression though.

Issue #184, opened 2019-03-18
1.7.0 breaks SSR


Betree:
After updating from 1.6.2 to 1.7.0, we get the following error:

```
ReferenceError: window is not defined
    at Object.<anonymous> (/home/user/Workspace/Projects/opencollective/opencollective-frontend/node_modules/react-window/dist/index.cjs.js:78:3)
    at Module._compile (internal/modules/cjs/loader.js:722:30)
    at Module._compile (/home/user/Workspace/Projects/opencollective/opencollective-frontend/node_modules/pirates/lib/index.js:83:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:733:10)
    at Object.newLoader [as .js] (/home/user/Workspace/Projects/opencollective/opencollective-frontend/node_modules/pirates/lib/index.js:88:7)
    at Module.load (internal/modules/cjs/loader.js:620:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:560:12)
    at Function.Module._load (internal/modules/cjs/loader.js:552:3)
    at Module.require (internal/modules/cjs/loader.js:658:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.reactWindow (webpack:/external "@babel/runtime/regenerator":1:1)
    at __webpack_require__ (webpack:///webpack/bootstrap:27:1)
    at Module.componentsInputTypeCountryJs (/home/user/Workspace/Projects/opencollective/opencollective-frontend/src/.next/server/static/development/pages/collective.js:18682:53)
    at __webpack_require__ (webpack:///webpack/bootstrap:27:1)
    at Module.componentsInputFieldJs (/home/user/Workspace/Projects/opencollective/opencollective-frontend/src/.next/server/static/development/pages/collective.js:17696:59)
    at __webpack_require__ (webpack:///webpack/bootstrap:27:1)
```

We use NextJS for pre-rendering, so `window` can be `undefined`.

I think the two lines causing this are:
- https://github.com/bvaughn/react-window/blob/master/src/createGridComponent.js#L135
- https://github.com/bvaughn/react-window/blob/master/src/createListComponent.js#L115

PR #185, opened 2019-03-18
Check for undefined window to support SSR

Betree:
Should fix https://github.com/bvaughn/react-window/issues/184


bvaughn:
Thanks

bvaughn:
Released in 1.7.1

PR #188, opened 2019-03-19
Add persistedIndexes prop to createListComponent

rssfrncs:
Couldn't find a contributing guide so not sure what the policies are here! 😁

This PR attempts to add a prop named "persistIndexes" to createListComponent. Any indexes specified will be rendered (persisted) even when they are out of the current window. It will enable the popular use-case "sticky header" as asked for in #55 with a relatively small change.

Example https://codesandbox.io/s/kmov2m4903?fontsize=14&view=preview,

```JSX
import { FixedSizeList as List } from 'react-window';
 
const Row = ({ index, style }) => (
  <div style={style} className={index === 0 ? "sticky" : null}>Row {index}</div>
);
 
const Example = () => (
  <List
    height={150}
    itemCount={1000}
    itemSize={35}
    width={300}
    // array of indexes to persist regardless of scroll position
    persistedIndexes={[0]}
  >
    {Row}
  </List>
);
```

```css 
.sticky {
  position: sticky !important;
  top: 0px;
}
```

This might very well be a simple and extremely naive implementation... 😊

### DIFF
--- a/src/createListComponent.js
+++ b/src/createListComponent.js
@@ -61,6 +61,7 @@ export type Props<T> = {|
   style?: Object,
   useIsScrolling: boolean,
   width: number | string,
+  persistedIndexes?: number[],
 |};
 
 type State = {|
@@ -261,6 +262,7 @@ export default function createListComponent({
         style,
         useIsScrolling,
         width,
+        persistedIndexes,
       } = this.props;
       const { isScrolling } = this.state;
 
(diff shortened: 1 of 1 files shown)


bvaughn:
I appreciate the PR!

To be honest though, I still don't think sticky items are a common enough use case for me to want to add them to the base library. I try to push back pretty strongly against features like this to help keep the library small and easily maintainable.

Here's an idea I had for how you could implement a component in userland that adds this behavior using the existing public API. Maybe you'd be interested in releasing something like this as a separate NPM package?

https://codesandbox.io/s/0mk3qwpl4l

rssfrncs:
@bvaughn totally reasonable and this example is great!

Issue #74, opened 2018-10-21
FixedSizeList padding


khlieng:
Is there a way to have padding at the top and bottom of a list? If I do `10px 0` on the outer div I get `20px` of padding at the bottom, I can make it look proper by adding a `10px` top margin to the list items but that just shifts the windowing off-center.

bvaughn:
No, padding like that isn't really supported. If you want padding around the entire list, put the list inside of a padded container.

If you literally just want 10px

khlieng:
Ah ok, I wanted the padding inside the list like this https://jsfiddle.net/9wpd58mx/6/, I could just do it with a VariableSizeList.

ahtcx:
This can be fixed by adding `position: relative;` to the list, should I make a PR with a fix @bvaughn ?

henryoliver:
Just subtract a given `offset` from the grid `height`:

```
const HEADER_OFFSET = 72;
...
return (
  <FixedSizeGrid
    ref={photoGridRef}
    width={width}
    height={height - HEADER_OFFSET}
    rowCount={rowCount}
    rowHeight={rowHeight}
    ...
```

PR #200, opened 2019-04-04
Allow paddingTop and paddingBottom in Lists

dariocravero:
Hi,

I know that in #74, @bvaughn said the way to add `paddingTop` and `paddingBottom` to a list would be through using a `VariableSizeList` and making the first and last elements pushers.

That seemed to work alright for a `Tree` component we implemented using `react-window` (which we'll hopefully open source at some point :)) but it was actually broken underneath.

The problem we faced was that in order to get the right size for the pushers we needed to increase the lists' size by two and then use an `itemSize` function that would determine the height. However, because that function caches the sizes aggressively, we were forced to put a `key` into the `List` to force it to re-render (couldn't find any other way) and recalculate the heights for each element.

The problem with that approach is that the focus is lost and the tree scrolls back to the top as soon as you select something in the middle.

If we don't use the `key`, then when the list grows in size, we have a gap in what should've been the last element but it isn't anymore.

I dug a bit on what it would mean to add the paddings to the list and unless something it didn't add all that much complexity to the code. It would be good if someone could sound the `scrollTo` part though.

What do you think?
Thanks!
Darío

Edit: I've released it here in the meantime https://www.npmjs.com/package/@viewstools/react-window-padded-lists so we can use it internally

### DIFF
--- a/src/FixedSizeList.js
+++ b/src/FixedSizeList.js
@@ -5,8 +5,8 @@ import createListComponent from './createListComponent';
 import type { Props, ScrollToAlign } from './createListComponent';
 
 const FixedSizeList = createListComponent({
-  getItemOffset: ({ itemSize, size }: Props<any>, index: number): number =>
-    index * ((itemSize: any): number),
+  getItemOffset: ({ itemSize, paddingTop, size }: Props<any>, index: number): number =>
+    paddingTop + index * ((itemSize: any): number),
 
   getItemSize: ({ itemSize, size }: Props<any>, index: number): number =>
     ((itemSize: any): number),
@@ -15,7 +15,7 @@ const FixedSizeList = createListComponent({
     ((itemSize: any): number) * itemCount,
 
   getOffsetForIndexAndAlignment: (
-    { direction, height, itemCount, itemSize, layout, width }: Props<any>,
+    { direction, height, itemCount, itemSize, layout, paddingTop, width }: Props<any>,
     index: number,
     align: ScrollToAlign,
     scrollOffset: number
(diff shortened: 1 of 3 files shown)


bvaughn:
Thanks for the PR. It's going to take me a while to review it. My backlog is slammed at the moment.

Couple of thoughts after an initial glance though:
* Grid would need similar padding left/right/top/bottom
* Needs tests
* Needs docs changes

dariocravero:
Nice! No worries. I'm super busy with stuff as well but didn't want to forget about opening the conversation at least.
If someone else wants to take in the todos you're suggesting, feel free to have a go at it. Maybe @khlieng (who commented on #74) might be interested? Otherwise I'll try to get to it at some later point.

bvaughn:
Cool. Thanks~

nihgwu:
@dariocravero I think you could use `resetAfterIndex` instead of changing the `key` to reset the caches

dariocravero:
@nihgwu I guess it would work. However, would it still mean that you have to add pushers at the top and bottom of the list?

nihgwu:
@dariocravero Yes, I just want to point out there is another way to stale the `itemSize` function

bvaughn:
Closing this due to inactivity.

We can pick back up the conversation if someone is willing to address [the initial PR feedback items](https://github.com/bvaughn/react-window/pull/200#issuecomment-479953254).

henryoliver:
Just subtract a given `offset` from the grid `height`:

```
const HEADER_OFFSET = 72;
...
return (
  <FixedSizeGrid
    ref={photoGridRef}
    width={width}
    height={height - HEADER_OFFSET}
    rowCount={rowCount}
    rowHeight={rowHeight}
    ...
```

PR #209, opened 2019-04-12
New scroll align mode: 'smart'

gaearon:
This adds implementation + tests. I can write the docs if you'd like them to be a part of the PR.

The purpose is to add a new scrolling mode called `smart`. We may make it default in v2 if we want.

It behaves like `auto` when the target is *within a single screen from the viewport boundaries*, and behaves like `center` otherwise. The rationale is that if something is far enough, there's no meaningful context to preserve — and therefore we might as well center it.

In particular, centering is helpful for cases when the next selected item is relatively random (such as with "find by DOM node" picker in DevTools). In that case, the `auto` behavior often jumps between the top and the bottom selected nodes because we're not necessarily scrolling in any particular direction. However, the `smart` behavior tries to center things unless we're already pretty close to them.

### Before

Screen Recording 2019-04-12 at 06 09 PM

### After

Screen Recording 2019-04-12 at 06 08 PM

### DIFF
--- a/src/FixedSizeGrid.js
+++ b/src/FixedSizeGrid.js
@@ -46,6 +46,14 @@ const FixedSizeGrid = createGridComponent({
         ((columnWidth: any): number)
     );
 
+    if (align === 'smart') {
+      if (scrollLeft >= minOffset - width && scrollLeft <= maxOffset + width) {
+        align = 'auto';
+      } else {
+        align = 'center';
+      }
+    }
+
     switch (align) {
       case 'start':
         return maxOffset;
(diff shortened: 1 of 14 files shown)


bvaughn:
Thanks Dan! I updated the docs site.

Published as v1.8.0

bvaughn:
> We may make it default in v2 if we want.

Had a laugh earlier tonight when I saw this comment. Version 2 has finally been released (though I didn't change the default) and it's only been 6 years 🫠

gaearon:
Made me laugh because I saw the PR in notifications, thought it’s some random discussion where I’m complaining about something, and it turns out it’s my own PR which I have zero recollection of.

bvaughn:
Happens to me often

PR #210, opened 2019-04-12
Change the pointer events value to be valid

oliviertassinari:
I have found the issue in https://validator.w3.org/nu/?doc=https%3A%2F%2Fnext.material-ui.com%2Fdemos%2Flists%2F. We have [a demo](https://next.material-ui.com/demos/lists/#virtualized-list) with this library.

Capture d’écran 2019-04-13 à 01 43 38

*(n°1 is an issue with Next.js)*

### DIFF
--- a/src/createListComponent.js
+++ b/src/createListComponent.js
@@ -326,7 +326,7 @@ export default function createListComponent({
           ref: innerRef,
           style: {
             height: isHorizontal ? '100%' : estimatedTotalSize,
-            pointerEvents: isScrolling ? 'none' : '',
+            pointerEvents: isScrolling ? 'none' : null,
             width: isHorizontal ? estimatedTotalSize : '100%',
           },
         })


src/createListComponent.js
```diff
@@ -326,7 +326,7 @@ export default function createListComponent({
           ref: innerRef,
           style: {
             height: isHorizontal ? '100%' : estimatedTotalSize,
-            pointerEvents: isScrolling ? 'none' : '',
+            pointerEvents: isScrolling ? 'none' : null,
```
bvaughn: Oh weird 😮I would have sworn the else condition was `undefined`

bvaughn:
The browser seems to drop an empty-string `pointer-events` style though from what I can see, so I don't know if this would actually have had any impact? Still, I'll publish a bugfix release shortly.

bvaughn:
v1.8.1

oliviertassinari:
I haven't noticed any impact. I aim at zero error reported by w3.org validator :).

PR #211, opened 2019-04-13
List components: add 'useIsVisible' prop

henryqdineen:
With `react-virtualized` we use the `isVisible` prop from `rowRenderer` to conditionally render a "lightweight" version of the row. This feature doesn't exist in `react-window` but it seemed straightforward to add.

I attempted to implement this in our app by getting the visible index data from `onItemsRendered` and passing that along to `itemData` but it did not seem like it would be the most performant solution. 

I'm open to feedback about whether this feature is useful or if there are other patterns that would be preferable.

### DIFF
--- a/src/__tests__/FixedSizeList.js
+++ b/src/__tests__/FixedSizeList.js
@@ -870,4 +870,28 @@ describe('FixedSizeList', () => {
       );
     });
   });
+
+  describe('useIsVisible', () => {
+    it('should not pass an isVisible prop to children unless requested', () => {
+      ReactTestRenderer.create(<FixedSizeList {...defaultProps} />);
+      expect(itemRenderer.mock.calls[0][0].isVisible).toBeUndefined();
+    });
+
+    it('should pass an isVisible prop to children if requested', () => {
+      ReactTestRenderer.create(
+        <FixedSizeList {...defaultProps} useIsVisible />
+      );
+      expect(itemRenderer.mock.calls[0][0].isVisible).toBeDefined();
+    });
+
+    it('should pass an accurate isVisible prop to children', () => {
+      ReactTestRenderer.create(
+        <FixedSizeList {...defaultProps} useIsVisible />
+      );
+
+      expect(itemRenderer).toHaveBeenCalledTimes(7);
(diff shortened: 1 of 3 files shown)


JESii:
This looks as if it would solve our current requirements: we have multiple Lists on a page (one for each of several days) and we need to filter the list so that users can search for text inside the components which are displayed in the list. From a quick look at the commits, I see a range of visible items: is that going to work for our use case where visible and non-visible lines are scattered throughout the viewport? If so, this would be a terrific feature for us.

bvaughn:
> With react-virtualized we use the isVisible prop from rowRenderer to conditionally render a "lightweight" version of the row. This feature doesn't exist in react-window but it seemed straightforward to add.

I'm not sure this makes sense to me. The purpose of overscanning is to render a small number of additional rows in advance, so that if the user starts scrolling again quickly- we reduce the chance of them seeing empty space. If you render something _different_ for those advance rows, then we actually do _wasted_ work- since we need to throw away that temporary content and render "visible" content when the user starts scrolling. Seems like this would strictly _harm_ performance.

Am I looking at this wrong or overlooking a use case?

As I currently understand this feature, I don't think I want to add it, so I'm going to close this PR. Let's keep talking though and if I'm overlooking something, I'll give it another consideration!

I don't think I understand @JESii's use case either. Searching (at least if you're talking about the browser's built-in search) is not something that will work well with any virtualization library at the moment. That's why libraries like [js-worker-search](https://github.com/bvaughn/js-worker-search/) [js-search](and https://github.com/bvaughn/js-search/) exist.

JESii:
Hi, @bvaughn ... just to close the loop on your question. We want to be able to filter the list (not a really big list) and I thought a quick way would be to to this would be by just "skipping" the items that I wanted to filter out (I actually created a simple proof of concept [here](https://codesandbox.io/s/github/JESii/VariableSize-dynamicList)).
In the end, I just reload the table with the filtered results... much simpler.
In any event, thanks for a great plugin and for the references to other libraries which I'll review.

bvaughn:
Sweet. Glad you found a solution~

PR #212, opened 2019-04-14
export getScrollbarSize

nihgwu:
that would be useful if we need that helper and don't need to dupe the code again

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,3 +7,4 @@ export { default as FixedSizeList } from './FixedSizeList';
 
 export { default as areEqual } from './areEqual';
 export { default as shouldComponentUpdate } from './shouldComponentUpdate';
+export { getScrollbarSize } from './domHelpers';


bvaughn:
I understand the motivation behind this PR, but the method is only a couple bytes. I suggest just using the [dom-helpers](https://www.npmjs.com/package/dom-helpers) package if you need functionality like this. The method doesn't really have anything to do with windowing, and I don't want to be stuck supporting it or providing it if I decide to change the approach the library uses internally.

nihgwu:
> The method doesn't really have anything to do with windowing

The method is not needed for windowing, but is useful if you want to create components based on `react-window`, it just like `areEqual` and `shouldComponentUpdate` are utils for the consumers, I agree that it's better to use `dom-helpers` if we need those helpers

> and I don't want to be stuck supporting it or providing it if I decide to change the approach the library uses internally.

It's not necessary the implementation should be kept updated with the upsteam, it works as expected and that's it, also consumers can still import the method from other packages or implement their own ones

nihgwu:
I'm trying to migrate from `react-virtualized` to `react-window` for my table component, and there is `onScrollbarPresenceChange` to get the scrollbar's width before but not available in `react-window`, and I find the method from the source code and want to use it directly without adding another dependency or duplicating the code as you did in this package

bvaughn:
I'm sorry, but I'm not willing to add this export. The helper method is an internal implementation detail. It's not part of the public API.

nihgwu:
Never mind, thanks for your quick response 👍

bvaughn:
No problem 😄

PR #215, opened 2019-04-17
Fix scrolling issues in RTL in Chrome, Firefox and Edge

### LINKED ISSUE #159: RTL support is slightly broken for Safari and Edge
Based on the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft) it looks like Chrome is reports incorrect values for `scrollLeft` when the direction is "rtl":
> If the element's direction is rtl (right-to-left), then scrollLeft is 0 when the scrollbar is at its rightmost position (at the start of the scrolled content), and then increasingly negative as you scroll towards the end of the content.

Rather than reporting a range of <= 0, Chrome reports the distance in pixels measured from the left. (View the console output in [this repro case](https://qzqylo21zq.codesandbox.io/).)

I've tried to accommodate this broken behavior (fb15909 and dcb9c55) but the fix is kludgy.

It also doesn't fully work for Safari and Edge:
1. Go to [the demo page](https://react-window.now.sh/#/examples/list/fixed-size-rtl)
1. Scroll left a bit
1. Scroll back right

It looks like when the scroll position reaches the right, the `scrollLeft` value sometimes rolls over from 0 to a really high number, causing the list to blank out.

I am not very familiar with RTL layouts. Help would be appreciated.

### LINKED ISSUE #198: Synchronizing a grid and a list in RTL has unexpected behavior
In the sandbox linked to below, the top list has its scroll position set whenever the grid below is scrolled. If you change the direction props of both the list and the grid to `"ltr"`, all works fine.

If you use it as provided with the list and grid direction props set to `"rtl"` in Chrome the list doesn't appear to render anything, then when you scroll horizontally items do appear but only some of them and they appear to scroll the wrong way. In Firefox the list is initially rendered correctly but as soon as you start to scroll the grid horizontally the list items don't scroll and then disappear.

https://codesandbox.io/s/1077qx12n3

This issue is probably related to #159 .

MarkFalconbridge:
Hopefully this fixes #159 and fixes #198.

The code is essentially now keeping an internal normalized representation of scrollLeft which is used as scrollLeft used to be to determine the indices to be rendered. Anywhere that scrollLeft is reported out or received in, e.g. `onScroll` and `scrollTo` it reports and expects scrollLeft as the browser would represent it.

I've used the `normalize-scroll-left` library to determine what type of scrolling the current browser implements.

I've changed the ScrollToItem file to make it easier to see the changes in action hopefully proving the validity of the code changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,8 @@
     "flow-bin": "^0.80.0",
     "gh-pages": "^1.1.0",
     "lint-staged": "^7.0.5",
+    "normalize-scroll-left": "^0.1.2",
+    "path": "^0.12.7",
     "prettier": "^1.12.1",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "memoize-one": ">=3.1.1 <6"
+    "memoize-one": ">=3.1.1 <6",
+    "normalize-scroll-left": "^0.1.2"
   },
   "peerDependencies": {
     "react": "^15.0.0 || ^16.0.0",
@@ -94,7 +95,6 @@
     "flow-bin": "^0.80.0",
     "gh-pages": "^1.1.0",
     "lint-staged": "^7.0.5",
-    "normalize-scroll-left": "^0.1.2",
     "path": "^0.12.7",
     "prettier": "^1.12.1",
     "react": "^16.8.4",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,3 +1,4 @@
+import path from 'path';
 import babel from 'rollup-plugin-babel';
 import commonjs from 'rollup-plugin-commonjs';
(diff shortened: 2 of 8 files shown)


bvaughn:
This is a side project of mine. I will look when I get the time and have the energy 😅

RTL issuea are particularly unpleasant to deal with and I'm not really looking forward to sorting through the prs and issues about it.

MarkFalconbridge:
I completely understand. The fix I've written wasn't pleasant to write at all but I've tested it in all the scenarios I can think of and think I've got it covered.

An alternative solution would be to write some sort of shim to patch how browsers get and set `scrollLeft`, i.e. patching the HTMLElement prototype. The upside of this would be that the behavior would then be normalized and react-window's code could remain largely as is apart from removing any rtl specific code. The downside would be that anyone already trying to handle the various browser implementations of scrollLeft would see different behavior.

bvaughn:
Thanks for this PR!

I think I've found a simpler solution via PR #251 after looking at this and the related #159. Going to close this in favor of that.

Issue #217, opened 2019-04-18
w3.org states elements with `role='gridcell'` should be owned by an element with `role='row'`


MarkFalconbridge:
https://www.w3.org/TR/wai-aria-1.1/#grid states that elements with `role='gridcell'` should be owned by an element with `role='row'`.

With the current implementation of `FixedSizedGrid` or `VariableSizedGrid` I don't think this is possible as there aren't any elements surrounding a row of cells once rendered in the DOM.

Is this something that could be added so that full aria support becomes possible?

bvaughn:
Grouping grid cells into rows is not something I'm interested in doing. Adding the overhead of additional DOM element wrappers to performance sensitive areas like grid doesn't feel like a worthwhile tradeoff.

Is there a different aria role that would be better suited to the current DOM structure?

MarkFalconbridge:
Unfortunately not, I've definitely got a grid of data to present and according to that spec, elements with role `gridcell`, `columnheader`, or `rowheader` are owned by elements with role `row`, which are in turn owned by an element with role `rowgroup`, or `grid`. What this means in practice is that users who use assistive technologies may not get a full picture of how the data in the grid is structured.

thielium:
I can second that this is an issue for my team. 

The only way I know how to fulfill this requirement now is using `React.Children` to traverse the items passed into the inner element and using the items' props to group them into rows. Performance cost aside, there are problems with this approach, one being that it's not very idiomatic to have props travel up from the children into the parent like that. 

I have an [idea](https://github.com/thielium/react-window/commit/a2e7098adb965d7303635265c5c4b3552c6ee946) I'm not yet happy with that explores exposing a rendering mechanism for such purposes. It solves accessibility challenges, but it only partially enables sticky row / column handling, which I sort of feel like is in the same category as this (i.e. cell decoration and grouping). I also am sympathetic to not wanting to expose too much of the inner workings to prevent the API from getting cluttered.

But I'm very interested in this particular issue's resolution. I don't want to be pedantic, but the language in the spec that @MarkFalconbridge linked to is actually a little more forceful (emphasis not added): 

>Authors _**MUST**_ ensure elements with role gridcell, columnheader, or rowheader are owned by elements with role row, which are in turn owned by an element with role rowgroup, or grid.

Beyond just wrapping things with a `role="row"`, things like headers are pretty tough to achieve without some fine-grained controls. 

Happy to help on this one, if you find a solution but don't find the time, @bvaugh.

[...]

PR #218, opened 2019-04-19
Adds range rendering

thielium:
In certain situations, consumers benefit from being able to control how virtualized items are placed, such as when managing aria attributes (where the DOM hierarchy determines accessibility; see
https://www.w3.org/TR/wai-aria-1.1/#grid).

In this pull request, one could fulfill the accessibility requirement with a `cellRangeRenderer` function, such as the example below: 

```js
  const accessibleCellRangeRenderer = ({
    rowStartIndex,
    rowStopIndex,
    columnStartIndex,
    columnStopIndex,
    childFactory,
  }) => {
    const rows = [];
    for (let rowIndex = rowStartIndex; rowIndex <= rowStopIndex; rowIndex++) {
      const rowChildren = [];
      for (
        let columnIndex = columnStartIndex;
        columnIndex <= columnStopIndex;
        columnIndex++
      ) {
        const cell = childFactory({ rowIndex, columnIndex });
        rowChildren.push(cell);
      }
      rows.push(<div role="row">{rowChildren}</div>);
    }

    return rows;
  };
```

Should address https://github.com/bvaughn/react-window/issues/217

### DIFF
--- a/src/__tests__/FixedSizeGrid.js
+++ b/src/__tests__/FixedSizeGrid.js
@@ -4,6 +4,7 @@ import ReactTestRenderer from 'react-test-renderer';
 import ReactTestUtils from 'react-dom/test-utils';
 import { FixedSizeGrid } from '..';
 import * as domHelpers from '../domHelpers';
+import { defaultCellRangeRenderer } from '../createGridComponent';
 
 const findScrollContainer = rendered => rendered.root.children[0].children[0];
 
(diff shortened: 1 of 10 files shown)


bvaughn:
This code change seems solid.

I need to think on it more before I am willing to commit to merging it though. Getting rid of the range renderer prop felt like a _nice_ change from react-virtualized to react-window. Re-adding it would add extra function calls to relatively hot code.

thielium:
Closing -- no time to work on this, and the iron's no longer hot.

PR #229, opened 2019-04-30
rename overscan props for Gird with deprecation

### LINKED ISSUE #226: overscanRowsCount vs overscanRowCount
it's `overscanRowCount` in `react-virtualized`, but now it's `overscanRowsCount`, while we have `rowCount`, I'm not a native speaker so I'm just wondering which is proper?  Right now I have to map `overscanRowCount` to `overscanRowsCount` https://github.com/Autodesk/react-base-table/pull/19/files#diff-4177f614bd864f6344923aed07c2edeaR91

nihgwu:
close #226

### DIFF
--- a/src/__tests__/FixedSizeGrid.js
+++ b/src/__tests__/FixedSizeGrid.js
@@ -289,20 +289,43 @@ describe('FixedSizeGrid', () => {
         spyOn(console, 'warn');
 
         const renderer = ReactTestRenderer.create(
+          <FixedSizeGrid {...defaultProps} overscanCount={1} />
+        );
+        expect(console.warn).toHaveBeenCalledTimes(1);
+        expect(console.warn).toHaveBeenLastCalledWith(
+          'The overscanCount prop has been deprecated. ' +
+            'Please use the overscanColumnCount and overscanRowCount props instead.'
+        );
+
+        renderer.update(<FixedSizeGrid {...defaultProps} overscanCount={1} />);
+
+        // But it should only warn once.
+        expect(console.warn).toHaveBeenCalledTimes(1);
+      });
+
+      it('should warn about deprecated overscanRowsCount or overscanColumnsCount prop', () => {
+        spyOn(console, 'warn');
+
+        const renderer = ReactTestRenderer.create(
           <FixedSizeGrid
             {...defaultProps}
-            overscanCount={1}
             overscanRowsCount={1}
             overscanColumnsCount={1}
           />
         );
         expect(console.warn).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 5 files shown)


src/__tests__/FixedSizeGrid.js
```diff
@@ -289,12 +289,17 @@ describe('FixedSizeGrid', () => {
...
+            overscanCount={1}
+            overscanRowsCount={1}
+            overscanColumnsCount={1}
+          />
         );
         expect(console.warn).toHaveBeenCalledTimes(1);
         expect(console.warn).toHaveBeenLastCalledWith(
-          'The overscanCount prop has been deprecated. ' +
-            'Please use the overscanColumnsCount and overscanRowsCount props instead.'
+          'The overscanCount, overscanColumnsCount and overscanRowsCount props have been deprecated. ' +
```
bvaughn: I think this message may be a bit confusing for people who were unaware of the old `overscanCount` prop. Seems like it would be better to split these into two warnings?

nihgwu: OK, will do

nihgwu:
Hey @bvaughn , I'd like to ask for your help for an off-topic problem, https://github.com/Autodesk/react-base-table/pull/22, do you have any idea to avoid a second rendering to solve the problem, that is if I know it's a pure client rendering I could just use the real scrollbar size in construction, or I have to update it in `cDM`

I also found that in `react-virtualized` https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/Grid.js#L610 this statement is wrong, and it won't work actually, as in client hydration the component's `constructor` `render` will be called again and the scrollbar size is measured there, so this block won't be hit at all

I noticed this problem after I migrated to `react-window`, it wasn't  a problem as I always do a second rendering because I relied on `onScrollbarPresenceChange` from `Grid` before, but now I calculate it by myself for migrating to `react-window`

website/src/routes/api/FixedSizeGrid.js
```diff
@@ -300,13 +312,25 @@ const PROPS = [
...
+  {
+    description: (
+      <p>
+        <strong className={styles.DeprecatedProp}>
+          This property has been deprecated.
+        </strong>{' '}
+        Please use the <code>overscanRowCount</code> property instead.
+      </p>
+    ),
+    name: 'overscanRowsCount',
```
bvaughn: I think you forgot to rename `overscanRowsCount` -> `overscanRowCount` for the pre-existing one. 😁 I'll do it!

bvaughn:
Released in 1.8.2 - https://github.com/bvaughn/react-window/blob/master/CHANGELOG.md#182

Issue #242, opened 2019-05-20
[feature request] Support passing props to outer element.


ricokahler:
(at work, we've been getting into the habit of using SBAR to communicate and I've noticed it's a great format for feature requests)

## Situation

There is no way to pass additional props to the outer `div` of any of the react-window components.

## Background

I'm trying to mimic the "arrow key stepper" functionality from react-virtualized. In order to do this, I would like add the prop `tabIndex` to the order-most `div` of the `FixedSizeList` component from `react-window` so that when the element is focused, I can key keyboard events using `onKeyDown`.

## Assessment

From my perspective, it doesn't seem like allowing users to pass additional props to the outer div would break any existing functionality (though I may be missing something).

See https://github.com/bvaughn/react-window/pull/243 for more details.

That PR adds the requested functionality and all the test still pass 🎉

## Recommendations

Allow users to pass props through that first outer div. Other popular component libraries like material-ui do this by convention.

TrySound:
You may pass any component to `outerElementType`. That component may render div or another element like you want. To pass props from outside you may use react context.

bvaughn:
There are a few ways of passing data down with the existing API. I don't like the pattern of spreading rest props for a library like react-window. (Maybe I can go into the "why" at some point when I'm less rushed for time than I am this morning.)

Anyway... I think there are existing ways to do what you're going for within the current API:

### Passing props through

Check out [the docs](https://react-window.now.sh/#/api/FixedSizeList) for the `outerElementType` proper @TrySound mentions:
> **`outerElementType: React$ElementType = "div"`**
Tag name passed to document.createElement to create the outer container element. This is an advanced property; in most cases, the default ("div") should be used.

This one probably most closely resembles your PR. [Here is an example sandbox.](https://codesandbox.io/s/bvaughnreactwindow-fixedsizelistvertical-ciqp2)

### Sharing data

If you're just sharing data then I suggest using the [`itemData` prop](https://react-window.now.sh/#/examples/list/memoized-list-items) or the React context API.

oliviertassinari:
We have faced the same issue to implement this demo: https://material-ui.com/components/autocomplete/#virtualization.

We have used the following approach:
```jsx
const ListboxComponent = React.forwardRef(function ListboxComponent(props, ref) {
  const { ...other } = props;

  const outerElementType = React.useMemo(() => {
    return React.forwardRef(
      (props2, ref2) => <div ref={ref2} {...props2} {...other} />,
    );
[...]

PR #243, opened 2019-05-20
Allow pass through props

ricokahler:
This PR attempts to allow pass through props. See https://github.com/bvaughn/react-window/issues/242 for more context.

There are a few flow errors I don't know how to fix and I'm not sure if some of the props I added to the destructure are all necessary.

Thanks for the awesome library and let me know what you think!


bvaughn:
Thanks for the PR! Please see [my comment](https://github.com/bvaughn/react-window/issues/242#issuecomment-494023419) for my reasoning about why I don't want to make this change at the present time.

PR #250, opened 2019-05-23
[Issue 244] List flickers on iOS during rubber band overscrolling

blenoski:
Added logic to determine whether or not we have overscrolled
vertical scrolling boundaries of list.

Updated list vertical scrolling handler to skip rerenders
if we have overscrolled.


EyMaddis:
The same applies for horizontal scrolling, but is ignored in this PR

bvaughn:
Thanks for this PR!

I'm incorporating the ideas from this along with RTL PRs #162 and #215 into PR #251.

PR #261, opened 2019-06-11
Enable implicit inexact object lint

TrySound:


src/VariableSizeGrid.js
```diff
@@ -19,7 +19,7 @@ type ItemMetadata = {|
   offset: number,
   size: number,
 |};
-type ItemMetadataMap = { [index: number]: ItemMetadata };
+type ItemMetadataMap = { [index: number]: ItemMetadata, ... };
```
bvaughn: I'm unfamiliar with this "..." syntax, and my attempts to find it in the docs are failing. Pointers?

bvaughn: Looks like CI is unfamiliar with it also 😆 

Screen Shot 2019-06-11 at 8 27 11 AM

TrySound: Explicit inexact object. Part of transition to exact by default.
https://github.com/facebook/flow/releases/tag/v0.101.0

TrySound:
Looks like it's not yet supported by babel-eslint

src/VariableSizeGrid.js
TrySound: Will add it later then.

bvaughn: Gotcha. Thanks. I don't think they've published the 101 notes internally yet. I've only read v100.

PR #262, opened 2019-06-12
Implicit inexact object lint

TrySound:
Ref https://github.com/bvaughn/react-window/pull/261

I found the version of babel-eslint which supports explicit inexact objects


TrySound:
Looks like now we have to fight with `react-scripts`. This is why I don't like such "wrapper" tools.

Issue #166, opened 2019-03-08
Fixed grid headers with CSS position sticky


apazzolini:
Hi @bvaughn,

I'm sure you're really tired of this request, but per [your comment here](https://github.com/bvaughn/react-window/issues/69#issuecomment-427364423) I think I might have a lightweight-enough solution for you to consider.

This solution keeps browser-native scrolling, doesn't require any ScrollSync-esque behavior, and only requires a few changes to the `createGridComponent` render method. However, IE11 is not supported. I think this is OK as long as it's explicitly called out in the documentation. 

You can see a demo here: https://codesandbox.io/s/xjzrzmxv8q

This is what I have locally as a prototype - if you're interested in a PR for this, there's definitely some more work to be done as several things are hardcoded. I can do the work here, but would appreciate some guidance.

I totally understand if you don't want to add this - in the worst case, this issue can at least document the approach for others.

An overview of my changes:

First, when generating `items`, we want to skip row == 0 and column == 0 (the Math.max() calls). We also push into leftStickyItems and topStickyItems for the visible range. Those are the only changes here.

```js
const items = [];
const topStickyItems = [];
const leftStickyItems = [];

if (columnCount > 0 && rowCount) {
  for (
    let columnIndex = Math.max(1, columnStartIndex); // Skip column 0
    columnIndex <= columnStopIndex;
    columnIndex++
  ) {
    topStickyItems.push(
      createElement(children, {
        columnIndex,
        data: itemData,
        isScrolling: useIsScrolling ? isScrolling : undefined,
        key: itemKey({ columnIndex, data: itemData, rowIndex: 0 }),
        rowIndex: 0,
        style: this._getItemStyle(0, columnIndex),
      })
    );
  }

  for (
    let rowIndex = Math.max(1, rowStartIndex); // Skip row 0
    rowIndex <= rowStopIndex;
    rowIndex++
  ) {
    // I'm leveraging this already existing loop, but this is probably
    // better as its own loop for clarity.
    leftStickyItems.push(
      createElement(children, {
        columnIndex: 0,
        data: itemData,
        isScrolling: useIsScrolling ? isScrolling : undefined,
        key: itemKey({ columnIndex: 0, data: itemData, rowIndex }),
        rowIndex,
        style: this._getItemStyle(rowIndex, 0),
      })
    );

    for (
      let columnIndex = Math.max(1, columnStartIndex); // Skip column 0
      columnIndex <= columnStopIndex;
      columnIndex++
    ) {
[...]

PR #265, opened 2019-06-13
Added Sticky Grid Header demo from #166 to README

john-osullivan:
If you think this demo is clean enough to share with new users, then it'd be nice to have it surfaced right on the README

https://github.com/bvaughn/react-window/issues/166


bvaughn:
This example depends on some custom build of `react-window` (https://github.com/apazzolini/react-window)

john-osullivan:
Just realized my error, my b!  Thanks for addressing this so quickly, much appreciated.

PR #267, opened 2019-06-14
Prevent toggling between min and max scroll offsets at end of grid

### LINKED ISSUE #257: FixedSizeGrid.scrollToItem sometimes scrolls the item only half into view
If the to scroll to is at the bottom of the grid and is already in view, calling scrollToItem actually scrolls the grid so that the item is only partially visible.

Here (https://codesandbox.io/s/bvaughnreactwindow-scrollingtoagriditem-d8cc4) is a sandbox to illustrate the issue. If you manually scroll the grid so that cell "r9, c5" is fully visible then click the "scroll to row 9 column 5" button you'll see that the grid scrolls so that you can only see half of the cell. (alternatively just keep clicking the button and you'll see the cell scroll into view and then scroll halfway out of view)

MarkFalconbridge:
Fixes #257 

Prevents toggling between the calculated minOffset and maxOffset when minOffset is greater than maxOffset owing to the scrollbar size.

### DIFF
--- a/src/FixedSizeGrid.js
+++ b/src/FixedSizeGrid.js
@@ -117,6 +117,8 @@ const FixedSizeGrid = createGridComponent({
       default:
         if (scrollTop >= minOffset && scrollTop <= maxOffset) {
           return scrollTop;
+          // Because we only take into account the scrollbar size when calculating minOffset
+          // this value can be larger than maxOffset when at the end of the list
         } else if (minOffset > maxOffset) {
           return minOffset;
         } else if (scrollTop - minOffset < maxOffset - scrollTop) {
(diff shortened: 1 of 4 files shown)


src/FixedSizeGrid.js
```diff
@@ -65,6 +65,8 @@ const FixedSizeGrid = createGridComponent({
       default:
         if (scrollLeft >= minOffset && scrollLeft <= maxOffset) {
           return scrollLeft;
+        } else if (minOffset > maxOffset) {
+          return minOffset;
```
bvaughn: I think this needs an explanatory comment, because it looks like a bug at first glance 😄

MarkFalconbridge: Fair point! Now added.

bvaughn: Looks like you added other places but missed this one  😆

bvaughn:
Published in 1.8.3

PR #268, opened 2019-06-14
Add support for RTL in Edge

MarkFalconbridge:
For #159 

This fix adds support for non-bleeding edge Edge which uses a different positive scrollLeft when in RTL compared to the positive scrollLeft that Chrome uses.

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -3,7 +3,7 @@
 import memoizeOne from 'memoize-one';
 import { createElement, PureComponent } from 'react';
 import { cancelTimeout, requestTimeout } from './timer';
-import { getScrollbarSize, isRTLOffsetNegative } from './domHelpers';
+import { getScrollbarSize, getRTLOffsetType } from './domHelpers';
 
 import type { TimeoutID } from './timer';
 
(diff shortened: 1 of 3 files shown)


src/createGridComponent.js
```diff
@@ -364,12 +364,17 @@ export default function createGridComponent({
...
-          if (isNegative) {
-            outerRef.scrollLeft = -scrollLeft;
-          } else {
-            const { clientWidth, scrollWidth } = outerRef;
-            outerRef.scrollLeft = scrollWidth - clientWidth - scrollLeft;
+          switch (getRTLOffsetType()) {
+            case 'negative':
+              outerRef.scrollLeft = -scrollLeft;
+              break;
+            case 'reverse':
```
bvaughn: These labels are confusing to me. Especially calling a non-spec-compliant behavior "default". I wonder if there is a more meaningful name we could pick?

So I think we're differentiating between:
* spec-compliant browsers: 0 ... -N
* Chrome: N ... 0
* non-Chromium Edge: 0 ... N

Maybe the following labels would be a bit more meaningful?
* negative: (0 ... -N)
* positive-descending: (N ... 0)
* positive-ascending: (0 ... N)

MarkFalconbridge:
I'm happy with the renames.

bvaughn:
Published in 1.8.3

PR #272, opened 2019-06-22
Add ItemMeasurer prop to DynamicSizeList

natew:
Not familiar with flow so didn't add proper typing yet.


bvaughn:
I don't think this is a direction I want to go with the component. Thanks for the PR though!

PR #274, opened 2019-06-24
FixedSizeList fixes for onItemsRendered & partially-visible items

### LINKED ISSUE #270: Off-by-one error in onItemsRendered reports visibleStopIndex that's not visible
My app uses react-window like a carousel: there's only one item visible at a time, and `size === itemSize`. I was surprised to see that `onItemsRendered` reports that `visibleStopIndex` is always one larger than `visibleStartIndex`, even though that's not what's actually being shown to the user. I traced the issue to this code:

https://github.com/bvaughn/react-window/blob/aa71ef99990eb3e559a0afafb3a09fa2f90eefc4/src/FixedSizeList.js#L77-L96

If `size` is an even multiple of `itemSize` (including my case where they're the same), and if the item is snapped to the edge (I use `scroll-snap-type` and `scroll-snap-align` to guarantee this), then the code above reports a `stopIndex` that's not actually a visible item.

I'm happy to PR a fix, but I wanted advice because this would technically be a breaking change for users who are relying on current behavior, e.g. they're subtracting one from the stop index, perhaps assuming that `visibleStopIndex` is intended to be that way, like the way C++ standard library collections always expect `end` to be one index more than the index of the last element in the collection.

In theory a less-breaking change could be to make *all* `onItemsRendered` calls return one more than the real stop index, even if items aren't snapped to the edge or if the `size` isn't an even multiple of `itemSize`.  At least then the behavior wouldn't change depending on the ratio of the item to container size.  I admittedly don't like this idea-- seems like a bad tradeoff to make the API permanently more confusing to avoid a major revision bump.

Got ideas for how this should be addressed?

justingrant:
This PR fixes a few related issues with `FixedSizeList`:
* Updates test snapshots to correctly calculate expected item indices for all `onItemsRendered` calls for `FixedSizeList`.  This was done with pen-and-paper and manual calculation to ensure that the existing implementation didn't influence expected results. Most of the changes involved reducing  `visibleStopIndex` and `overscanStopIndex` by one index because of #270.
* Adds tests for partially-visible items, which exposed a few more bugs beyond #270.
* Reports `visibleStopIndex` is correctly in `onItemsRendered`. Previously, `visibleStopIndex` was one item higher than the last visible index.  Fixes #270.
* Ensures that partially-visible items are correctly handled in `onItemsRendered` and `scrollToItem`.
* Fixes the `align: 'center'` option of `scrollToItem` in cases where the viewport is near the beginning or end of the list where the existing method of calculating the centered offset (averaging `minOffset` and `maxOffset`) didn't produce a correct result.
* Changes the behavior of the `align: 'auto'` option of `scrollToItem` to handle the case where items don't fit evenly in the viewport. Now, when scrolling forward the partially-visible item will always be shown hanging off the end of the viewport. Exception: at the end of the list, there's no room to show the partial at the end of the viewport so it's shown at the beginning instead. Previously, forward scrolls would hang the partial item off the top of the viewport, which IMHO looked really weird when there wasn't an obvious reason for the visual weirdness (like being at the end of the list). Note that partially-visible items were broken in several ways previously, so I don't think this is a breaking change-- because no one could have used the current implementation successfully with `size % itemSize !== 0`.  Also note that this change doesn't affect backwards scrolls because the partial (if any) was already shown at the bottom when scrolling up in the current implementation.
* Updates the docs with notes relating to the changes above. 
* Updates the docs for `overscanCount` to show a default value of `2` (which matches react-window's source code). Previously, the default value shown was `1` which didn't match the code.  We could go the other way and change the default in the code to match the docs (which would make the default faster!) but I'd worry about making this change without an explicit breaking change because it would affect user-visible behavior during fast scrolls. 
* In `getOffsetForIndexAndAlignment`, simplified the conditional `scrollOffset - minOffset < maxOffset - scrollOffset` to `scrollOffset < minOffset` which returns equivalent results in that context but is more readable. 

In theory there is a breaking change in this PR: fixing `visibleStopIndex`/`overscanStopIndex` may break clients who are depending on the broken behavior, e.g. if they assume that the `*StopIndex` values were supposed to be exclusive.  @bvaughn - I dunno if this is worth a major version bump or not.

Note that this PR only fixes `FixedSizeList`. Looking at the source for `VariableSizeList` and the `*Grid` components, it's likely that they share many of the same issues.  In a perfect world I'd have fixed those too, but this PR took more time than I expected so I didn't have time to fix other components. Hopefully someone can use changes in this PR as inspiration for fixing those other components too!

### DIFF
--- a/src/FixedSizeList.js
+++ b/src/FixedSizeList.js
@@ -23,12 +23,13 @@ const FixedSizeList = createListComponent({
     // TODO Deprecate direction "horizontal"
     const isHorizontal = direction === 'horizontal' || layout === 'horizontal';
     const size = (((isHorizontal ? width : height): any): number);
-    const maxOffset = Math.max(
+    const lastViewportOffset = Math.max(
       0,
-      Math.min(
-        itemCount * ((itemSize: any): number) - size,
-        index * ((itemSize: any): number)
-      )
+      itemCount * ((itemSize: any): number) - size
+    );
+    const maxOffset = Math.min(
+      lastViewportOffset,
+      index * ((itemSize: any): number)
     );
     const minOffset = Math.max(
       0,
@@ -51,14 +52,35 @@ const FixedSizeList = createListComponent({
         return maxOffset;
       case 'end':
         return minOffset;
-      case 'center':
-        return Math.round(minOffset + (maxOffset - minOffset) / 2);
+      case 'center': {
+        // "Centered" offset is usually the average of the min and max
+        // offsets. But near the beginning or end of the list, this math
+        // doesn't produce the actual closest-to-center offset, so we
+        // override.
+        const centered = Math.round(minOffset + (maxOffset - minOffset) / 2);
+        if (centered < Math.ceil(size / 2)) {
+          return 0; // near the beginning
+        } else if (centered > lastViewportOffset + Math.floor(size / 2)) {
+          return lastViewportOffset; // near the end
+        } else {
+          return centered;
+        }
+      }
       case 'auto':
       default:
         if (scrollOffset >= minOffset && scrollOffset <= maxOffset) {
           return scrollOffset;
-        } else if (scrollOffset - minOffset < maxOffset - scrollOffset) {
-          return minOffset;
+        } else if (scrollOffset < minOffset) {
+          // For auto alignment, if size isn't an even multiple of itemSize,
+          // always put the partial item at the end of the viewport because it
+          // looks better than at the beginning. Exception: don't do this if the
+          // viewport is already scrolled to the end.
+          const remainder = size % ((itemSize: any): number);
+          if (minOffset + remainder < lastViewportOffset) {
+            return minOffset + remainder;
+          } else {
+            return minOffset;
+          }
         } else {
           return maxOffset;
         }
@@ -83,14 +105,17 @@ const FixedSizeList = createListComponent({
     const isHorizontal = direction === 'horizontal' || layout === 'horizontal';
     const offset = startIndex * ((itemSize: any): number);
     const size = (((isHorizontal ? width : height): any): number);
+    // How far before the scrollOffset does the first visible item start?
+    // Will be zero if scrollOffset is an item boundary.
+    const startingPartialSize = scrollOffset - offset;
+    const visibleItems = Math.ceil(
+      (size + startingPartialSize) / ((itemSize: any): number)
+    );
     return Math.max(
       0,
       Math.min(
         itemCount - 1,
-        startIndex +
-          Math.floor(
-            (size + (scrollOffset - offset)) / ((itemSize: any): number)
-          )
+        startIndex + visibleItems - 1 // -1 is because stop index is inclusive
       )
     );
   },
--- a/src/__tests__/FixedSizeList.js
+++ b/src/__tests__/FixedSizeList.js
@@ -77,7 +77,7 @@ describe('FixedSizeList', () => {
 
   it('should render a list of rows', () => {
     ReactTestRenderer.create(<FixedSizeList {...defaultProps} />);
-    expect(itemRenderer).toHaveBeenCalledTimes(7);
+    expect(itemRenderer).toHaveBeenCalledTimes(6);
     expect(onItemsRendered.mock.calls).toMatchSnapshot();
   });
 
@@ -85,7 +85,7 @@ describe('FixedSizeList', () => {
     ReactTestRenderer.create(
       <FixedSizeList {...defaultProps} layout="horizontal" />
     );
-    expect(itemRenderer).toHaveBeenCalledTimes(5);
+    expect(itemRenderer).toHaveBeenCalledTimes(4);
     expect(onItemsRendered.mock.calls).toMatchSnapshot();
   });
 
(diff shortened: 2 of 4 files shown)


justingrant:
From a brief look at `VariableSizeList` and the grid components, it looks like most/all of the problems will affect all of the other components too.  The changes in this PR can probably be applied unchanged (or almost unchanged) to the other 3 components, although updating the tests is non-trivial.  Would it be bad to fix only `FixedSizeList` for now and add a new issue with a Help Wanted label to fix the others?

bvaughn:
I think it's probably not good to merge a change that results in different
behavior between the four components.

That's not to say you need to fix them all, but it might be a while before
I get around to it.

bvaughn:
I think it's probably not a good idea to merge a change that results in different behavior between the list/grid components.

That's not to say you need to fix them all, but it might be a while before I get around to it. I might try to spend some time on it this weekend, but no promises on when I'll be able to finish it 😄

src/FixedSizeList.js
```diff
@@ -51,14 +52,35 @@ const FixedSizeList = createListComponent({
...
       case 'auto':
       default:
         if (scrollOffset >= minOffset && scrollOffset <= maxOffset) {
           return scrollOffset;
-        } else if (scrollOffset - minOffset < maxOffset - scrollOffset) {
-          return minOffset;
+        } else if (scrollOffset < minOffset) {
+          // For auto alignment, if size isn't an even multiple of itemSize,
+          // always put the partial item at the end of the viewport because it
+          // looks better than at the beginning. Exception: don't do this if the
```
bvaughn: Not sure I understand the reasoning behind this added complexity. Why does it "look better"?

bvaughn: I think I disagree with this change. Probably going to remove it. It breaks the specification of "auto" scroll behavior, by over-scrolling in the case when you're scrolling from a lower to higher index.

src/FixedSizeList.js
```diff
@@ -51,14 +52,35 @@ const FixedSizeList = createListComponent({
         return maxOffset;
       case 'end':
         return minOffset;
-      case 'center':
-        return Math.round(minOffset + (maxOffset - minOffset) / 2);
+      case 'center': {
+        // "Centered" offset is usually the average of the min and max
+        // offsets. But near the beginning or end of the list, this math
+        // doesn't produce the actual closest-to-center offset, so we
+        // override.
```
bvaughn: This change looks good 👍

src/FixedSizeList.js
justingrant: Imagine a list with `itemSize=30` and `size=100`.  Current behavior when scrolling forward (lower to higher index) with `align=auto` will look like `|--|------|------|--xx--|`, where 'xx' is the desired item and each `-` represents 5px.

I'm suggesting that `|------|------|--xx--|--|` is a better way to handle the case when items don't fit exactly into the viewport, because people generally read top-to-bottom (or left-to-right for horizontal LTR, or right-to-left for horiz RTL).  Users are used to seeing partial items at the end of a list, but seeing them at the start of a list is a somewhat jarring user experience that I was hoping to avoid. 

It's something of a judgement call.  I don't feel very strongly about this one way or the other, esp. because my current use of react-window doesn't trigger this case.  If you take it out I won't be offended. ;-)

BTW, I've been using the relatively-new CSS setting `scroll-snap-type` and it really improves UX for react-window by enforcing whole-item scrolling, so you're not left with partial items visible.

bvaughn: Yeah, I think I understood the suggestion just not the reasoning behind it 😄 thanks for elaborating~

> BTW, I've been using the relatively-new CSS setting `scroll-snap-type` and it really improves UX for react-window by enforcing whole-item scrolling, so you're not left with partial items visible.

Ah, that's interesting. I haven't played with the two yet. I can see how it might be a UX improvement!

bvaughn:
I'm going to be making some changes in this branch btw. (Just letting you know so that we don't conflict.)

src/FixedSizeList.js
justingrant: Yep `scroll-snap-type` and `scroll-snap-align` have been big (and very easy) UX wins. Another nice one has been  and it massively improves UX. 

I've also had good results with `scroll-behavior: smooth;` which makes `scrollToItem` look really slick as the item slides gracefully into position. The only caveat with this one is that if you have an initial scroll offset that's large, then it smooth-scrolls on mount which can take 500-1000 msecs of items speeding by before it gets to the right item. IMHO this looks bad (and delays initial user interaction) so I've had to add this style using a `useEffect` after mounting.

It's amazing how much better browsers in general (and CSS in particular) have gotten in the last 5 years.  The surface area is huge though-- hard to keep up with!

justingrant:
No worries. I wasn't planning to make any changes until I see what cool stuff you're doing. ;-)

justingrant:
BTW, @bvaughn one line of code I didn't fully understand was this one: 

https://github.com/bvaughn/react-window/blob/c71baadbbf14a038d320e0ebb3a8757dcff24e05/src/FixedSizeList.js#L60

`scrollOffset - minOffset < maxOffset - scrollOffset` simplifies to `scrollOffset < (maxOffset - minOffset) / 2`.  In English: "scroll offset is smaller than the average of min and max offsets". 

But looking at the context of that code...

https://github.com/bvaughn/react-window/blob/c71baadbbf14a038d320e0ebb3a8757dcff24e05/src/FixedSizeList.js#L56-L64

... the condition above it (`scrollOffset >= minOffset && scrollOffset <= maxOffset`) already handled the case where `scrollOffset` was between `minOffset` and `maxOffset`, so isn't `scrollOffset - minOffset < maxOffset - scrollOffset` equivalent to `scrollOffset < minOffset`?

Mostly I didn't understand the reason for the additional complexity of that comparison, which made me wonder if I was missing something more fundamental.

src/__tests__/FixedSizeList.js
```diff
@@ -421,13 +421,59 @@ describe('FixedSizeList', () => {
...
+      // No need to scroll again; item 9 is already visible.
+      // Because there's no scrolling, it won't call onItemsRendered.
+      rendered.getInstance().scrollToItem(9, 'auto');
+      // Scroll to near the end. #96 will be shown as partial.
+      rendered.getInstance().scrollToItem(99, 'auto');
+      // Scroll back to show #96 fully. This will cause #99 to be shown as a
+      // partial. Because #96 was already shown previously as a partial, all
+      // props of the onItemsRendered will be the same. This means that even
+      // though a scroll happened in the DOM, the onItemsRendered event is
+      // not triggered. IMHO this is probably a bug, but it's one we can fix
```
bvaughn: This would be working as per the spec. Why do you consider this a bug? `onItemsRendered` should only be called when the rendered indicates change.

bvaughn:
> so isn't `scrollOffset - minOffset < maxOffset - scrollOffset` equivalent to `scrollOffset < minOffset`?

Yeah I think you're right that this could be simplified :)

bvaughn:
So the good news is, I ported the new `FixedSizeList` tests over to `VariableSizeList` and they all passed without any changes. So maybe the bugs were just related to the fixed sized implementation.

bvaughn:
Squashed+merged in 51a24c4. Will be releasing shortly.

bvaughn:
Fixes published as 1.8.4

Thanks for your help

src/__tests__/FixedSizeList.js
justingrant: Yep, you're right. At the time I wrote the comment above I thought that `onItemsRendered` was expected to be called when an item changed from being partially to fully visible. Sorry I forgot to update that comment!

justingrant:
Very cool. Thanks @bvaughn!

Ironically, after I filed this PR I realized that I needed to know which items are partially vs. fully visible (e.g. to switch selection in one list after another list scrolls.. but not until the new item scrolls fully into view). So I had to switch from `onItemsRendered` to `onScroll` instead.

Regardless, it was fun to be able to help make react-window better. Good library BTW... very useful. Thanks so much for making it great!

Issue #232, opened 2019-04-30
Support UMD Release


stunaz:

bvaughn:
I would accept a PR that adds a UMD build, if you'd like to submit one.

stunaz:
I wish I could help.Could you flag it as `help wanted`, hopefully somebody would help.

bvaughn:
Honestly, it's not something I care about supporting. If you have a use case for it,  PRs are welcome though 😄

bvaughn:
Cleaning up issues today. Don't intend to actively support this one so I'm going to close it.

bvaughn:
v1.8.5 adds a UMD build

stunaz:
@bvaughn thanks a lot :-)

PR #281, opened 2019-07-05
support for the UMD package format

emmanueltouzery:
I saw in https://github.com/bvaughn/react-window/issues/232 that you'd consider a PR to support UMD package generation, and it would be very important for us to have a UMD package of react-window honestly.

So here's a PR. I tested the generated UMD with our app and it seems to work just fine. There are two open questions maybe:

1. I tested by copying the generated JS files in our app. We just need to be sure that the files will indeed be in the NPM package. I believe they'll be, but...

2. If you want to just provide UMD support with the minimal extra dependencies, maybe you don't want to "uglify" (minimize) the output. For our purpose we don't need a minimized UMD, but some people may expect it. So I made two commits, the uglify support is in the second commit. So you can take it in, or not. I checked and formik for instance ships minified UMDs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,8 @@
     "rollup": "^1.4.1",
     "rollup-plugin-babel": "^4.3.2",
     "rollup-plugin-commonjs": "^9.2.1",
-    "rollup-plugin-node-resolve": "^4.0.1"
+    "rollup-plugin-node-resolve": "^4.0.1",
+    "rollup-plugin-replace": "^2.2.0",
+    "rollup-plugin-terser": "^5.1.0"
   }
 }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,5 +1,7 @@
 import path from 'path';
 import babel from 'rollup-plugin-babel';
+import replace from 'rollup-plugin-replace';
+import { terser } from 'rollup-plugin-terser';
 import commonjs from 'rollup-plugin-commonjs';
 import nodeResolve from 'rollup-plugin-node-resolve';
 
@@ -8,6 +10,16 @@ import pkg from './package.json';
 const input = './src/index.js';
 
 const external = id => !id.startsWith('.') && !path.isAbsolute(id);
+
+const babelConfigEsModules = babel({
+  runtimeHelpers: true,
+  plugins: [['@babel/transform-runtime', { useESModules: true }]],
+  sourceMaps: true,
+});
+
+const umdGlobals = {
+  react: 'React',
+};
 
 export default [
   {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -61,7 +63,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "memoize-one": ">=3.1.1 <6"
+    "memoize-one": ">=3.1.1 <6",
+    "rollup-plugin-replace": "^2.2.0"
```
bvaughn: This should be a dev dependency, not a dependency

rollup.config.js
```diff
@@ -1,7 +1,9 @@
 import path from 'path';
 import babel from 'rollup-plugin-babel';
+import replace from 'rollup-plugin-replace';
 import commonjs from 'rollup-plugin-commonjs';
 import nodeResolve from 'rollup-plugin-node-resolve';
+import { uglify } from 'rollup-plugin-uglify';
```
TrySound: Use terser please

rollup.config.js
```diff
@@ -47,4 +49,58 @@ export default [
...
+    plugins: [
+      babel({
+        runtimeHelpers: true,
+        plugins: [['@babel/transform-runtime', { useESModules: true }]],
+        sourceMaps: true,
+      }),
+      nodeResolve(),
+      commonjs(),
+      replace({
+        "process.env.NODE_ENV": "'development'"
```
TrySound: JSON.stringify('development')

rollup.config.js
```diff
@@ -47,4 +49,58 @@ export default [
...
+      sourcemap: true,
+      name: 'ReactWindow',
+      globals: {
+        'react': 'React'
+      }
+    },
+    external: ['react'],
+    plugins: [
+      babel({
+        runtimeHelpers: true,
```
TrySound: Reuse esm babel config

rollup.config.js
```diff
@@ -47,4 +49,58 @@ export default [
...
+
+  {
+    input,
+    output: {
+      file: pkg.umddev,
+      format: 'umd',
+      sourcemap: true,
+      name: 'ReactWindow',
+      globals: {
+        'react': 'React'
```
TrySound: Define globals separately and reuse

rollup.config.js
```diff
@@ -47,4 +49,58 @@ export default [
...
+      }
+    },
+    external: ['react'],
```
TrySound: Object.keys(globals)

package.json
```diff
@@ -37,6 +37,8 @@
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
   "files": ["dist", "src/*.js"],
+  "umddev": "dist/index-dev.umd.js",
+  "umdprod": "dist/index-prod.umd.js",
```
TrySound: Not sure about this. Maybe hardcode in config?

emmanueltouzery:
ok, all clear! I'll resubmit with your suggestions + the yarn lint failures.

TrySound:
Just push in this branch

emmanueltouzery:
I pushed a new version!

bvaughn:
Thanks for the code review, @TrySound.

Published 1.8.5

PR #286, opened 2019-07-09
Fix umd global for react-dom

trenthogan:
Before:
Screen Shot 2019-07-09 at 1 23 33 PM

After:
Screen Shot 2019-07-09 at 1 24 55 PM

Issue is not on master branch as that doesn't appear to be using 
`import { findDOMNode } from 'react-dom';`


bvaughn:
Thanks

trenthogan:
👍

PR #321, opened 2019-09-04
Version 2 [WIP]

### LINKED ISSUE #302: Version 2 changes
This is an umbrella issue to share my plans for the upcoming version 2.0 release of `react-window`.

**Feedback is appreciated** and will be taken into consideration. I ask that you be understanding if I'm not able to accommodate all requests. I'll try to do my best to weigh feature requests against bundle size, runtime performance, and maintenance concerns.

I expect that upgrading from version 1 to 2 **may require substantial code changes**, many of which will not be possible to automate with code mods. Because of this, particularly for application code, I would advise against upgrading existing code unless you have a strong reason (e.g. you need support for dynamic size content).

I am also going to go ahead and pin the current docs to the domain [react-window-v1.now.sh](https://react-window-v1.now.sh/) so they will not be lost when I update for version 2.

---
##### Table of contents:
* Support fewer components
* Use render props API
* No more horizontal lists support
* Only grids support RTL
* Changes to `onScroll` callback timing
* Other changes/deprecations
---

### Fewer components

One way to help manage complexity is to **reduce the number of components the library supports**. I am currently planning on supporting only the following component types in version 2:

* `SimpleList` (previously [`FixedSizeList`](https://react-window.now.sh/#/examples/list/fixed-size))
  * This highly optimized list component should be used when row heights are fixed (and known ahead of time).
* `List` (previously [`DynamicSizeList`](https://react-window-next.now.sh/#/examples/list/dynamic-size))
  * This list should be used for dynamically sized content (e.g. chat, newsfeed). It requires the [`ResizeObserver`](https://wicg.github.io/ResizeObserver) API (or polyfill).
* `Grid` (previously [`VariableSizeGrid`](https://react-window.now.sh/#/examples/grid/variable-size))
  * This component should be used for tabular data (e.g. spreadsheets) that should be virtualized along both vertical and horizontal axis. It supports variable size rows and columns, but does not support automatically measuring and updating their sizes.

---
### Render props

One of the major changes from `react-virtualized` to `react-window` was the decision to treat `children` as React elements (e.g. `React.createElement(children, props))`) rather than as [render props](https://reactjs.org/docs/render-props.html) (e.g. `children(props)`).

There were a couple of motivations for doing this:
* React provides built-in solutions for memoization (e.g. [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo), [`useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo), [`shouldComponentUpdate`](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-shouldcomponentupdate)) so I didn't need to implement my own caching abstraction for item renderers.
* APIs like hooks and suspense "just work" inside of item renderers.
* Keys can be managed by `react-window` without requiring the render prop to pass them along (and without requiring `cloneElement` calls).

Unfortunately there were also some downsides:
* Inline item renderers incur a high cost. Because their "type" (the function definition) gets recreated each time the parent component renders, React deeply unmounts and remounts their rendered tree. This means that docs need to teach people not to use them even though they're often more convenient.
* Because inline functions couldn't be used to close over local scope, it was more complicated for item renderers to share state with parents, requiring [APIs like `itemData` and a custom `areEqual` comparison export](https://react-window.now.sh/#/examples/list/memoized-list-items).

After taking the above pros and cons into consideration, **I've decided to convert to a render props approach** for `react-window` as well. This means that complicated examples [like this](https://react-window.now.sh/#/examples/list/memoized-list-items) can be re-written more easily:

```js
const Example = ({ height, items, toggleItemActive, width }) => (
  <List
    height={height}
    itemCount={items.length}
    itemRenderer={({ index, key, style }) => {
      const item = items[index];
      return (
        <div key={key} onClick={() => toggleItemActive(index)} style={style}>
          {item.label} is {item.isActive ? "active" : "inactive"}
        </div>
      );
    }}
    itemSize={35}
    width={width}
  />
);
```

---
### No more horizontal list support

Previously, list components supported both horizontal and vertical layout modes. In order to simplify implementation and maintenance, and because the overwhelmingly common use case is vertical lists, I will be removing support for `layout="horizontal"`.

---
### RTL support

Grid components will continue to support `direction="RTL"`, but lists will not (since they will only support a vertical layout). This tradeoff is being made to enable lists to be smaller and easier to maintain.

---
### `onItemsRendered` and `onScroll` callback changes
List and grid components currently support `onItemsRendered` and `onScroll` callback props. These callbacks are called during the commit phase (after the list or grid has completed rendering). This can be useful in that it is always safe to perform a side effect (like analytics logging) in response to these callbacks, but it has a downside as well: any scroll synchronized updates must be done in a second ("cascading") render.

Version 2 will make a change to the `onScroll` callback to address this. **The `onScroll` callback will be called during the event's dispatch cycle** so that any update will be batched (by React) with the list or grid's own update.

**The `onItemsRendered` callback will be replaced with a `onItemsDisplayed` prop**, although it will continue to be called during the commit cycle. This change is being made to enable the list component to more aggressively optimize render performance by pre-rendering at idle priority and make use of experimental APIs like [display locking](https://www.chromestatus.com/feature/4613920211861504).

---
### Other props changes/deprecations

There are several pending deprecations (with DEV warnings) which will be removed:
* `innerTagName` and `outerTagName` for all list and grid components. (Use `innerElementType` and `outerElementType` instead.)
* `overscanCount`, `overscanColumnsCount`, and `overscanRowsCount` for grid component. (Use `overscanColumnCount` and `overscanRowCount` instead.)
* `overscanCount` will be removed for list components, in favor of a dynamic overscanning approach.
* "horizontal" and "vertical" values for `direction`. (These were moved to `layout`, but they will be removed entirely in version 2.)
* The `itemData` prop (and the corresponding `data` prop passed to item renderers) will be removed because the change to a render prop API no longer make this necessary.
* The `useIsScrolling` prop (and the corresponding `isScrolling` prop passed to item renderers) will be removed because the changes to pre-rendering and display locking would make this more expensive to implement.
* Scroll align parameters will change slightly. The previously named "auto" will now be named "minimal". The new default value will be "smart" (rather than "auto").

Note that some of the above deprecated props may not still be relevant given the other planned changes, but I'm listing them here anyway for completeness sake.

bvaughn:
Resolves #302

- [ ] Version 2 API
  - [x] `SimpleList` (previously [`FixedSizeList`](https://react-window.now.sh/#/examples/list/fixed-size))
  - [ ] `List` (previously [`DynamicSizeList`](https://react-window-next.now.sh/#/examples/list/dynamic-size))
  - [ ] `Grid` (previously [`VariableSizeGrid`](https://react-window.now.sh/#/examples/grid/variable-size))
- [ ] Update documentation

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -8,6 +8,5 @@
 
 [options]
 include_warnings=true
-suppress_type=$FlowFixMe
 
 [strict]
--- a/.flowconfig
+++ b/.flowconfig
@@ -8,5 +8,6 @@
 
 [options]
 include_warnings=true
+suppress_type=$FlowFixMe
 
 [strict]
--- a/package.json
+++ b/package.json
@@ -65,7 +65,8 @@
   },
   "peerDependencies": {
     "react": "^15.0.0 || ^16.0.0",
-    "react-dom": "^15.0.0 || ^16.0.0"
+    "react-dom": "^15.0.0 || ^16.0.0",
+    "scheduler": "^0.9.0"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
--- a/src/SimpleList.js
+++ b/src/SimpleList.js
@@ -2,7 +2,7 @@
 
 import memoizeOne from 'memoize-one';
 import { createElement, PureComponent } from 'react';
-// $FlowFixMe Cannot import flushSync because there is no flushSync export in react-dom.
+// $FlowFixMe
 import { flushSync as maybeFlushSync } from 'react-dom';
 import { cancelTimeout, requestTimeout } from './timer';
 
--- a/src/SimpleList.js
+++ b/src/SimpleList.js
@@ -1,0 +1,411 @@
+// @flow
+
+import memoizeOne from 'memoize-one';
+import { createElement, PureComponent } from 'react';
+import { cancelTimeout, requestTimeout } from './timer';
+
+import type { TimeoutID } from './timer';
+
+export type ScrollToAlign = 'center' | 'minimum' | 'smart' | 'start' | 'end';
+
+type Style = {
+  [string]: mixed,
+};
+
+type itemRendererFn = ({|
+  index: number,
+  key: string | number,
+  style: Style,
+|}) => React$Node;
+
+type onItemsDisplayedFn = (startIndex: number, stopIndex: number) => void;
+
+type ScrollEvent = SyntheticEvent<HTMLElement>;
+
+type ItemStyleCache = { [index: number]: Style };
+
+type OuterProps = {|
+  children: React$Node,
+  className: string | void,
+  onScroll: ScrollEvent => void,
+  style: Style,
+|};
+
+type InnerProps = {|
+  children: React$Node,
+  style: Style,
+|};
+
+export type Props = {|
+  className?: string,
+  height: number,
+  initialScrollOffset?: number,
+  innerRef?: any,
+  innerElementType?: string | React$AbstractComponent<InnerProps, any>,
+  itemCount: number,
+  itemKey?: (index: number) => string | number,
+  itemRenderer: itemRendererFn,
+  itemSize: number,
+  onItemsDisplayed?: onItemsDisplayedFn,
+  onScroll?: ScrollEvent => void,
+  outerRef?: any,
+  outerElementType?: string | React$AbstractComponent<OuterProps, any>,
+  style?: Object,
+  width: number | string,
+|};
+
+type State = {|
+  scrollTop: number,
+  scrollUpdateWasRequested: boolean,
+  startIndex: number,
+  stopIndex: number,
+|};
+
+const defaultItemKey = (index: number) => index;
+
+const RESET_POINTER_EVENTS_DEBOUNCE_INTERVAL = 150;
+
+export default class List extends PureComponent<Props, State> {
+  _innerRef: ?HTMLElement;
+  _outerRef: ?HTMLElement;
+  _resetPointerEventsTimeoutID: TimeoutID | null = null;
+
+  // Always use explicit constructor for React components.
+  // It produces less code after transpilation. (#26)
+  // eslint-disable-next-line no-useless-constructor
+  constructor(props: Props) {
+    super(props);
+
+    const scrollTop =
+      typeof this.props.initialScrollOffset === 'number'
+        ? this.props.initialScrollOffset
+        : 0;
+
+    const [startIndex, stopIndex] = this._getRangeToRender(scrollTop);
+
+    this.state = {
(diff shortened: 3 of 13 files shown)


bvaughn:
Not sure why Flow is failing on CI but not for me locally.

nihgwu:
Really looking forward the new components, I'm planning to switch to use List instead of Grid for the next version of BaseTable, I chose Grid instead of List to implement the BaseTable at first because the List doesn't support horizontal scrolling.

I've worked out a prototype and it's really promising, especially the change on `onScroll`, now I can get `scrollLeft` for [fixed mode(horizontal scrollable)](https://autodesk.github.io/react-base-table/examples/fixed/) without any workaround, but there is still a limitation there,  can we add an extra prop like `itemWidth` or `innerWidth` to set the inner element's width, and the default value is `100%`, although I can change the style directly, but that will break the memoization.

Another concern is about the `key` from the `itemRenderer`, I think it's me suggesting you to support callback instead of using `index` directly, but now as we switched to renderer, I think we could just remove `itemKey`, and let the users to set the key for item.

bvaughn:
> Another concern is about the `key` from the `itemRenderer`, I think it's me suggesting you to support callback instead of using `index` directly, but now as we switched to renderer, I think we could just remove `itemKey`, and let the users to set the key for item.

Hm, I think we could remove the `key` attribute from the `SimpleList` renderer, and rely on users to attach it themselves. They may forget, although in that case React would warn them.

I wouldn't want to remove it from the `List` renderer though, since we also may want to cache other info on that same key (e.g. offset, size) in which case we'd want a stronger guarantee that the keys would never mismatch.

I think this PR is probably going to be put on hold anyway as I will probably end up trying to build more of this into react core where we can do more and more efficiently.

nihgwu:
Then what about my first question? I understand you consideration the list should not be scrollable horizontally, but from my experience we need that feature in a lot of cases, btw I was talking about  SimpleList above actually and use List for short

bvaughn:
Lists can be horizontally scrollable if you want to make them. Overriding width style should not break your memoization. (This is true of v1 and v2.) I'd like to keep this PR discussion centered around the new API if possible.

zmitry:
Why not children as callback ?
``` 
<SimpleList>{({style},index)=>{...}}</SimpleList>
```
this way it's much more convenient.

HenrikBechmann:
@bvaughn "Lists can be horizontally scrollable if you want to make them." 

I *really* hope that's true! My app depends on it. 

Flexbox?

muhammedmagdi:
@bvaughn Any update on this? Would v2 be released soon or I have to download this branch manually for now?

PR #331, opened 2019-09-12
pass start and end indexes to outer element

interstella-5555:
Hello there!

I've implemented sticky columns and rows in an alternative way by filtering these columns/rows out and rendering them in the outer element instead with `position: sticky;`.
At this point it's working just fine but they are not windowed, all of the sticky cells are rendered. To achieve windowing I need to pass `columnStartIndex`, `columnStopIndex`, `rowStartIndex`, `rowStopIndex` to outer element which I did with my pull request. It doesn't produce any additional renders because at the point where these values may change, other props on outer element would change also.

I attach a rough demo of my implementation of the grid as well as a code playground for this POC for reference. The indexes would be used in `StickyCells` component and they would be implemented as just a simple `if` conditional for the cells.

Let me know what you think, thanks!

demo: https://1r94e.csb.app/
sandbox: https://codesandbox.io/embed/jovial-monad-1r94e

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -50,6 +50,10 @@ type OuterProps = {|
   children: React$Node,
   className: string | void,
   onScroll: ScrollEvent => void,
+  columnStartIndex: number,
+  columnStopIndex: number,
+  rowStartIndex: number,
+  rowStopIndex: number,
   style: {
     [string]: mixed,
   },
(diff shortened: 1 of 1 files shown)


artemirq:
@kwypchlo Thanks for sharing your implementation of sticky items using CSS Grid. It's works nice and I don't know any other universal solutions, which allow to pin rows/columns in any side and any count. So, I used your solution with some architectural changes, but main idea is the same. Thanks again

bvaughn:
Hm. I *think* maybe I'd be open to accepting this change, but we'd want to do it for Grid and List components, and I'd probably want to just hold off until v2 because I'm trying to avoid thinking about or making any more changes to v1 until v2 is out.

kwypchlo:
closing due to inactivity

Issue #333, opened 2019-09-15
[DynamicSizeList] Incorrect List Dimension When Size of Item Data is Reduced


jnishiyama:
I'm using branch `issues/6`, and I am experiencing a problem similar to #289 but not quite the same. Here's my best attempt at explaining the issue (codesandbox: https://codesandbox.io/s/bugreact-window-dynamic-size-list-djegq)

Let's say you have `n` items in your `itemData` and you scroll past `l` items. If `itemData` is reduced to `m` items where `m < l` then the height/width of the list will be incorrect.

I am not very well-versed with the code in `react-window`, but from what I can tell, the issue stems from this calculation in `DynamicSizeList.js`:

```
const getEstimatedTotalSize = (
  { itemCount }: Props<any>,
  {
    itemSizeMap,
    estimatedItemSize,
    lastMeasuredIndex,
    totalMeasuredSize,
  }: InstanceProps
) =>
  totalMeasuredSize + (itemCount - lastMeasuredIndex - 1) * estimatedItemSize;
```

When `getEstimatedTotalSize` is called after scrolling and reducing `itemCount`, `(itemCount - lastMeasuredIndex - 1) * estimatedItemSize;` will likely be negative.

When `getEstimatedTotalSize` is negative, the height/width set on the innerElement will remain as the last `estimatedTotalSize` value because you can't set a negative height. From createListComponent.js: `height: isHorizontal ? '100%' : estimatedTotalSize`. 

To fix this, I would propose just returning `itemCount * estimatedItemSize` when `totalMeasuredSize + (itemCount - lastMeasuredIndex - 1) * estimatedItemSize` is negative.

I will make a PR with this code change.

Freak613:
While the issue got stuck, here is ridiculous hack to update height directly:
```javascript
const maxHeight = 300;
const itemHeight = 30;

<List height={Math.min(maxHeight, itemHeight * items.length)}  ...>
```

OmkoBass:
> While the issue got stuck, here is ridiculous hack to update height directly:
> 
> ```js
> const maxHeight = 300;
> const itemHeight = 30;
> 
> <List height={Math.min(maxHeight, itemHeight * items.length)}  ...>
> ```

Didn't fix my issue. Anyone got an idea? As soon as i resize the list it just cuts itself in half.

OmkoBass:
Actually fixed my issue. `Math.min(maxHeight, itemHeight * items.length)` GOES TO itemSize not to height.

PR #334, opened 2019-09-15
[DynamicSizeList] fix for #333 - incorrect list size due to reduction in itemCount

jnishiyama:
Resolves Issue #333.

* Adds check for negative return in `getEstimatedTotalSize` in DynamicSizeList.js.

### DIFF
--- a/src/DynamicSizeList.js
+++ b/src/DynamicSizeList.js
@@ -124,15 +124,15 @@ const getEstimatedTotalSize = (
   }: InstanceProps
 ) => {
   const simpleEstimatedTotalSize = itemCount * estimatedItemSize;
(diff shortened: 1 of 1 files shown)


daviderenger:
Any news on getting this in a making a new release? I want to use the dynamic list for a filter list where number of items shrink as the user types. Right now the scrollable area keeps its biggest size forever. This fix seems to the job.

daviderenger:
Hi,

There is a case left where this is not working. If you have a long list, scroll down atleast one window height nad then do something to lower the amount if items. In my case a text filter it stays scrolled down and the overall height is wrong.

image


If I stay on top or close to top it works as expected.
I´ll do some digging and see if I can find what it is.

PR #350, opened 2019-10-18
Move defaultItemKey to defaultProps

Eddie-CooRo:
Move defaultItemKey to Component.defaultProps instead of setting it's default while destructuring in the render function

### DIFF
--- a/src/createListComponent.js
+++ b/src/createListComponent.js
@@ -66,7 +66,7 @@ export type Props<T> = {|
   innerTagName?: string, // deprecated
   itemCount: number,
   itemData: T,
-  itemKey?: (index: number, data: T) => any,
+  itemKey: (index: number, data: T) => any,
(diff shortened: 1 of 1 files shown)


bvaughn:
This change seems fine but it's not obvious to me that it's any better than what was there before, and since I'm working on v2 (very slowly) I'm going to just close this for now. I don't think I'll make any more changes to v1 components before v2 is released. Thanks though!

Eddie-CooRo:
Hey @bvaughn, thanks for your response.
This change was related to another improvement that I was trying to add in one of my projects. But unfortunately, I don't remember what the improvement was 😂.
What I remember, is that I needed to use the `defaultItemKey` in some other lifecycles, and because it's default value isn't in the `defaultProps`, I had to declare it's default value again while destructuring it in that lifecycle.

PR #391, opened 2019-12-15
better documentation

dkunzman77:
I think it would be helpful to have a couple of sentences in the beginning better describing what it does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,13 @@
 
 > React components for efficiently rendering large lists and tabular data
 
+> React window will only display a portion of your data but enable the user to scroll effortlessly through large datasets by updating what is displayed in the window.
+> It solves 3 important problems that are frequently encountered in software development.
+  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,13 @@
 
 > React components for efficiently rendering large lists and tabular data
 
+> React window will display only a portion of your data and enable the user to scroll effortlessly through large datasets.
+> React window solves 3 important problems that are frequently encountered in software development.
+  
+> 1) React window improves the responsiveness of your user interface by passing only a small portion of your data to the React reconciler.
+> 2) React window reduces the memory footprint of your application when displaying large datasets and prevents out of memory errors.
+> 3) React window reduces the need for your web services to support a paging interface.
```
bvaughn: This doesn't sound quite accurate. Windowing and data fetching are kind of separate issues, and often need to _both_ be done (using a helper lib like `react-window-infinite-loader`)

dkunzman77: It worked for us since are components were so large.  But I updated the pull request and removed the last sentence.

PR #411, opened 2020-02-02
Fixes issues in multi-window environment

rajjejosefsson:
When react-window is used in a multi-window environment, and React is running in the global/main window and renders to different child windows, react-window will not work as it should because it uses the global window and not the child window.

This PR fixes this issue by getting the current window from the `ref.parentNode.ownerDocument.defaultView` which has the proper reference to the current child window. If not we fallback to use the global window.

### DIFF
--- a/src/timer.js
+++ b/src/timer.js
@@ -18,7 +18,11 @@ export function cancelTimeout(timeoutID: TimeoutID, hostWindow: any) {
   hostWindow.cancelAnimationFrame(timeoutID.id);
 }
 
-export function requestTimeout(callback: Function, delay: number, hostWindow: any): TimeoutID {
+export function requestTimeout(
+  callback: Function,
+  delay: number,
+  hostWindow: any
+): TimeoutID {
   const start = now();
 
   function tick() {
--- a/src/timer.js
+++ b/src/timer.js
@@ -14,23 +14,23 @@ export type TimeoutID = {|
   id: AnimationFrameID,
 |};
 
-export function cancelTimeout(timeoutID: TimeoutID) {
-  cancelAnimationFrame(timeoutID.id);
+export function cancelTimeout(timeoutID: TimeoutID, hostWindow: any) {
+  hostWindow.cancelAnimationFrame(timeoutID.id);
 }
 
-export function requestTimeout(callback: Function, delay: number): TimeoutID {
+export function requestTimeout(callback: Function, delay: number, hostWindow: any): TimeoutID {
   const start = now();
 
   function tick() {
     if (now() - start >= delay) {
       callback.call(null);
     } else {
-      timeoutID.id = requestAnimationFrame(tick);
+      timeoutID.id = hostWindow.requestAnimationFrame(tick);
     }
   }
 
   const timeoutID: TimeoutID = {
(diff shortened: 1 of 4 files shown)


src/timer.js
```diff
@@ -14,23 +14,27 @@ export type TimeoutID = {|
   id: AnimationFrameID,
 |};
 
-export function cancelTimeout(timeoutID: TimeoutID) {
-  cancelAnimationFrame(timeoutID.id);
+export function cancelTimeout(timeoutID: TimeoutID, hostWindow: any) {
+  hostWindow.cancelAnimationFrame(timeoutID.id);
```
bvaughn: This looks like it should be unnecessary. Can you explain to me why this is necessary?

rajjejosefsson: Hi Brian, while running in a environment that are heavily based on working with multiple windows (iframe / webviews), then event listeners, DOM manipulation, measurements e.t.c. will mostly fail, because they are suddenly listening on/referring to an incorrect object. In this case what happens is that the pointersEvents never get reseted and always stays to be 'none'.

bvaughn: I'm still not sure I follow why this PR is necessary though given:
* Scrollbar size is an OS/browser thing and should return a consistent value regardless of root window or iframe, no? Maybe if the iframe is sandboxed, accessing `document.body` could be problematic I guess, in which case I guess I understand the `contentDocument` change.
* APIs like `cancelAnimationFrame` should not require a context. (JS running in an iframe shouldn't need to know it's running within an iframe to use them.) I've also assumed with methods like this, there's only a single global API shared between all frames/windows.

Can you explain how react-window breaks without these changes? Especially the RAF ones. I would not expect them to be necessary.

rajjejosefsson: My  app is running in one window and rendering it's output to a different window. The main window doesn't have any DOM output so requestAnimationFrame or setTimeout never fire in Chrome in this case. In order for those events to fire they must be called on the window object of the child window, which the current code doesn't allow.

Window object should be injectable and not depend on the global one.

Same issue have been addressed in [Autosizer - PR 1428](https://github.com/bvaughn/react-virtualized/pull/1428)

rajjejosefsson: Brian I understand your busy, but can we have this merged?

rajjejosefsson: @bvaughn I understand your busy, but can we have this merged?

smykhailov: hi @bvaughn this bug is pretty critical for our scenarios and we did already few similar changes in other libraries (Autosizer, Felajs, etc). Would it be possible to merge this PR, if you need better explanation about scenario it supports, I can prepare some codesnippet to demonstrate it. Thank you in advance.

decafdennis: @bvaughn When a window is obscured, Chromium stops firing `requestAnimationFrame` callbacks on that window. So, if you're rendering any part of your app in child window through a React portal (for example), then animation frames should be scheduled on that child window rather than the main window, despite being in the same context/renderer.

You can observe this by using DevTools, which effectively is the child window in the aforementioned scenario:

https://user-images.githubusercontent.com/93109/126084619-361d3af2-774b-46bb-9560-78765d3a7a47.mov

PR #421, opened 2020-02-15
Browsers without native ResizeObservable are now still able to use DynamicSizeList

tastyqbit:
Safari (and other non top 3 browsers) do not support ResizeObservable api natively and there was a lot of glitching due to that lack of support for changing heights based on width of the window.

This pull request fixes that by using @juggle/resize-observer if no native ResizeObservable exists

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "@babel/plugin-transform-runtime": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-flow": "^7.0.0",
+    "@juggle/resize-observer": "^3.0.2",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^9.0.0",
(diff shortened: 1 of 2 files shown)


bvaughn:
Thanks for the PR, but this doesn't look like a change I want to make. 🙂 The docs website for next version state pretty prominently that Resize Observer polyfill must be used if browser support for is required for e.g. Safari. I don't want to directly require this polyfill though. I think it should be opt-in. (Maybe you don't need to support such browsers. Maybe you use a different polyfill already. Etc)

tastyqbit:
Hey, that's for the feedback. Was there already a way to make it work with react-window that I just missed? Or are you still working on it?

bvaughn:
You should literally just be able to run a polyfill (in your application code) before using the component. I haven't been able to work on this library in a while though because I've been swamped with other things

PR #426, opened 2020-02-20
fix: don't unecessarily re-order elements

ronag:
Since the element layout uses absolut positioning the order does not matter. Avoid re-ordering.

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -445,6 +445,8 @@ export default function createGridComponent({
         }
       }
 
+      items.sort((a, b) => String(a.key).localeCompare(String(b.key)));
+
       // Read this value AFTER items have been created,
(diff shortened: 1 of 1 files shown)


deadbeef84:
This actually helps a lot when using css animations/transitions, since moving an element in the dom will reset its layout state preventing any animations unless you forcefully trigger a layout before applying the new style. When the dom-elements order is preserved, it's super easy to create masonry like animations using regular css transitions.

EarthlingDavey:
This may be cause unwanted accessibility side effects. If you have tab-able html elements like buttons in the rows. I think they will be highlighted in the order of the DOM, not the visual order from top down.
My suggestion, have this as something that is disabled by default. In the readme, caution developers about tab ordering and  reference to tabIndex as a possible solution.

Issue #234, opened 2019-05-05
[Feature request] Disable / control unmounting of already mounted rows


olee:
I would like to suggest an option to control the unmounting of already mounted rows.
This could give a performance boost to lists with a moderate amount of large rows but which are very expensive to render.
If properly handled (rows should use memo of course) it would only render rows the first time they are visible and then keep them mounted to be able to scroll back to them again without needing to render them again.

What do you think of this suggestion?
Might there be a way to do this with current react-window (or maybe react-virtualized?) right now?

bvaughn:
I don't think that sounds like something I want to support _directly_. It sounds like a reasonable use case, I just don't think I want to take on the extra complexity.

You could perhaps implement this behavior by supplying your own `innerElementType` that's stateful and maintains a list of rows it's created by inspecting its props:
https://github.com/bvaughn/react-window/blob/f2b2106cc828e9828cf2940411e5eab67f1d7aa6/src/createListComponent.js#L324-L331

But this would require some hacky duplication of this code:
https://github.com/bvaughn/react-window/blob/f2b2106cc828e9828cf2940411e5eab67f1d7aa6/src/createListComponent.js#L285-L298

Bnaya:
Out of curiosity:
What's the painful part in your use-case? the DOM creation or react rendering?
I think another approach is implementing DOM reuse, but it will require adding intermediate level of virtual keys.

olee:
Keeping react components mounted is pretty much the same as DOM reuse.
The initial rendering of new react nodes (ie. nodes that were added or have a new key) is the most expensive part of react rendering - subsequent updates are way faster.
This is why I would love to see a library supporting virtual scrolling, but at the same time keeping already rendered components mounted (because as long as they don't exceed a quite large amount, this should be faster).

PS: Oh yeah another option for "DOM reuse" might be a way to make the virtual scrolling **reuse the item keys**.
This should result in just their content being updated without making react disposing the old DOM elements and creating new ones.
However for this to work the items would have to be implemented in a way that ensures they do not totally rely on componentDidMount for some rendering logic.

Bnaya:
@olee 
I'm exploring possibilities when exposing the up-to-date overscan information to the keyProvider and cell components (its a patch on react-window),
https://codesandbox.io/s/github/Bnaya/sticky-eurovision/tree/iteration
Take a look at `src/implementations/WithReactWindowDOMReuse/WithReactWindowDOMReuse.tsx`

[...]

PR #475, opened 2020-06-06
Add virtualized indexes to `itemKey` function calls

jribeiro:
Currently only the real data index is passed to the `itemKey` functions for Grid and List components respectively.

While this is great to be able to get an unique key for each of the rendered items, when the user scrolls, the index will change causing the new key to mismatch the previous and React to unmount and re-mount each component and, as such, recreate the DOM node.

This PR adds a new parameter to the List component `itemKey` function call, `virtualizedIndex` which represents the row index relative to the first rendered row. Similarly, for Grids, there's two new properties: `virtualizedRowIndex`and `virtualizedColumnIndex`.

Depending on what we're optimising for, using these virtualized values instead can be a performance improvement and IMO it's quite good to have the additional option mostly when the code complexity remains pretty much the same.

Relates to: #234

### DIFF
--- a/src/__tests__/FixedSizeGrid.js
+++ b/src/__tests__/FixedSizeGrid.js
@@ -1049,12 +1049,42 @@ describe('FixedSizeGrid', () => {
         />
       );
       expect(itemKey).toHaveBeenCalledTimes(6);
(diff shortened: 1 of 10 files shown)


ORlGlN (approved):

fidelman:
any progress here?

PR #496, opened 2020-07-15
Upgrade fsevents from 1.2.3 to 1.2.13

devrelm:
`fsevents` v1.2.3 only has binaries available [up to Node 10](https://github.com/fsevents/fsevents/releases/tag/v1.2.3).

So when using Node 12.x+, it falls back to re-building `fsevents` from source. But that has its own set of issues that keeps it from working for some reason.

Upgrading `fsevents` to v1.2.13 appears to fix these issues.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -597,10 +597,6 @@ abab@^1.0.3, abab@^1.0.4:
   version "1.0.4"
   resolved "https://registry.yarnpkg.com/abab/-/abab-1.0.4.tgz#5faad9c2c07f60dd76770f71cf025b62a63cfd4e"
 
-abbrev@1:
-  version "1.1.1"
-  resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.1.tgz#f8f2c887ad10bf67f634f005b6987fed3179aac8"
-
 accepts@~1.3.4, accepts@~1.3.5:
   version "1.3.5"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -3781,22 +3762,17 @@ fs-extra@^4.0.2:
...
 
 fsevents@^1.0.0, fsevents@^1.1.2, fsevents@^1.1.3:
-  version "1.2.3"
-  resolved "https://registry.yarnpkg.com/fsevents/-/fsevents-1.2.3.tgz#08292982e7059f6674c93d8b829c1e8604979ac0"
+  version "1.2.13"
+  resolved "https://registry.yarnpkg.com/fsevents/-/fsevents-1.2.13.tgz#f325cb0455592428bcf11b383370ef70e3bfcc38"
+  integrity sha512-oWb1Z6mkHIskLzEJ/XWX0srkpkTQ7vaopMQkyaEIoq0fmtFVxOthb8cCxeT+p3ynTdkk/RZwbgG4brR5BeWECw==
   dependencies:
-    nan "^2.9.2"
-    node-pre-gyp "^0.9.0"
```
devrelm, line 3799: To generate the changes to the `yarn.lock`, I ultimately just removed these lines and ran `yarn install`.

Yarn then made the rest of the changes by:
* resolving the various `fsevents` dependencies to version 1.2.13
* removing resolutions for `node-pre-gyp` and its dependency graph
* adding resolutions for `bindings` and its dependencies graph

Issue #508, opened 2020-09-01
Keep last item of the list in the DOM to allow for focus-traps to properly cycle through list.


flasd:
Whenever I'm using a focus-trap for the virtualized list, when I hit the bottom edge of the list, the focus jumps to the first rendered element instead of the first element in the list.  

![ezgif com-video-to-gif (4)](https://user-images.githubusercontent.com/24439797/91884235-0f88ee80-ec5c-11ea-898a-673ffa8993d5.gif)

It there a way to keep the first and last elements of the list rendered to the DOM at all times?

PR #549, opened 2020-12-16
Add mustRender prop for list components

jclem:
This adds a `mustRender` prop for list components. An array of item indices can be provided, and react-window will render those items regardless of scroll position.

This feature is useful for cases where one may want to preserve focus on a component that the user may scroll off-screen. When it's scrolled back on-screen, one has to call `focus()` on it to maintain that focus state. Simply calling `focus()` on the elements when they're rendered is insufficient, as Safari does not support the `preventScroll` option that would prevent the browser from jumping immediately to some number of pixels past the newly-rendered focused component.

### DIFF
--- a/src/__tests__/FixedSizeList.js
+++ b/src/__tests__/FixedSizeList.js
@@ -98,6 +98,25 @@ describe('FixedSizeList', () => {
     expect(onItemsRendered.mock.calls).toMatchSnapshot();
   });
 
+  it('should throw when must-render items are out of bounds', () => {
+    spyOn(console, 'error');
+
+    expect(() => {
+      ReactTestRenderer.create(
+        <FixedSizeList
+          {...defaultProps}
+          mustRender={[defaultProps.itemCount]}
+        />
+      );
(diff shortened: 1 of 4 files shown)


EarthlingDavey:
Working for me with `VariableSizeList`.
Thanks for sharing your solution @jclem 👍

jclem:
It's been a little bit, so I'm going to go ahead and close this to clear it out of my personal backlog.

EarthlingDavey:
@bvaughn , please may you re-consider this pull request?

For rows, it allows DOM elements to retain focus, among other user interactions

And it allows react elements to retain state

Use cases could be when a row contains:

- form
- audio / video player
- slideshow / carousel
- anchor point
- modal

WofWca:
Closes https://github.com/bvaughn/react-window/pull/549.
Closes https://github.com/bvaughn/react-window/issues/795.
And this seems to be adding a way to fix https://github.com/bvaughn/react-window/issues/650 and https://github.com/bvaughn/react-window/issues/508.

PR #555, opened 2021-02-02
Add functionality to accurately calculate remaining height in VariableSizeList

KamranAsif:
## Problem:
Our `VariableSizeList` has very varied row heights. The default `estimatedItemSize` does not accurately represent the total height of the list.

This means users see a lot of jumpiness when scrolling down with the scrollbar. 

## Solution:
Adding `getEstimatedRemainingHeight` optional prop that returns an estimation of the remaining list height.
This function is passed the last measured index and expects to return the remaining height.

If accepted, I can add website documentation changes. I'm also open for a better name.

### DIFF
--- a/src/VariableSizeList.js
+++ b/src/VariableSizeList.js
@@ -8,6 +8,7 @@ const DEFAULT_ESTIMATED_ITEM_SIZE = 50;
 
 type VariableSizeProps = {|
   estimatedItemSize: number,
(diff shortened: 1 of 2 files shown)


omgoshjosh (approved):
LGTM

KamranAsif:
No longer need this; happy to let someone take this over and resubmit.

PR #600, opened 2021-10-21
increasing range of memoize-one to include 6.x

alexreardon:
`memoize-one@6.x` is an API compatible breaking change, with stronger TypeScript types

→ https://github.com/alexreardon/memoize-one/releases/tag/v6.0.0


alexreardon:
It is *absolutely amazing* that bumping to `memoize-one@6.0.0` occurred in issue #600

Issue #228, opened 2019-04-29
Calling scrollToItem() or scrollTo() triggers onScroll (callback) event twice, with inconsistent scrollUpdateWasRequestedProperty


tomasro27:
**Repro steps:**
1) create a callback function, and pass it as a `prop` for `onScroll` to `VariableSizeGrid` component.
2) Call `scrollToItem()` or `scrollTo()` 
3) Callback function gets called twice (in most cases.. see details below)
- First time, `scrollUpdateWasRequestedProperty` is `true`.
- Second time, `scrollUpdateWasRequestedProperty` is `false`.

Here is a **sandbox** with the repro:
https://codesandbox.io/s/4lq8xqwn14
1) Press `scrollToDecimalPX` button (using scrollTo with fractional part on PX value (500.34))
2) Check the console logs, and notice how the first one is true, and the second one is false.

Additional findings:
- **Fractional value** as an argument to `scrollTop` property in `scrollTo` function **always repro** (two calls to `onScroll` callback).
- **Integer value** as an argument to `scrollTop` property in `scrollTo` function makes **two calls** to  `onScroll` but **at least one out of ten times, it only calls it once.**
- `ScrollToItem` with **`align: auto`** or **`align: start`**  always calls `onScroll` **once** (as expected).
- `ScrollToItem` with **`align: end`** or  **`align: center`** always calls `onScroll` **twice**.
- `ScrollToItem` with **`align: smart`** calls `onScroll` **once or twice**.

Seems like the position in which we are scrolling to has an important role into whether `onScroll` gets called twice or once.

For context, I am trying to sync react-window `VariableSizeGrid` with another element B. I am trying to ignore the `onScroll` callback, but can't achieve that because of this.

bvaughn:
I don't think I have the time to look into this. If you'd like to submit a PR with unit tests I'll be happy to review and consider it.

telaoumatenyanis:
Hey @bvaughn.
I gave a look at this issue, here is what seems to happen:

The property `scrollTop` was originally a long and was changed into a double. I know chrome implemented the change, but I don't know if firefox did. I can't get a decimal scroll on firefox 67.

So for `Firefox`, as the value seems to be rounded down, the following check fail: 
https://github.com/bvaughn/react-window/blob/c71baadbbf14a038d320e0ebb3a8757dcff24e05/src/createGridComponent.js#L738-L741

The rounded value is different from the double one. It triggers a setState and thus call the callback.
[...]

PR #602, opened 2021-10-29
fix: Calling `scrollToItem` or `scrollTo` triggers onScroll twice

nicolasepiscopo:
Basically every time that the imperative scroll is being performed it would trigger a call to the `_onScroll` method, and that way there would be 2 events for each `scrollTo*`(one of them with `scrollUpdateWasRequested` as true, and the other as `false`). This fixes https://github.com/bvaughn/react-window/issues/228

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -195,6 +195,7 @@ export default function createGridComponent({
     _instanceProps: any = initInstanceProps(this.props, this);
     _resetIsScrollingTimeoutId: TimeoutID | null = null;
     _outerRef: ?HTMLDivElement;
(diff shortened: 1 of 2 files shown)


jasonLaster:
Hey @nicolasepiscopo, happy to take a look at the duplicate `onScroll` event. Would you be open to recording a replay with and without the fix? This way it will be easy to see how `_isPerformingImperativeScrolling` is being used.

https://docs.replay.io/docs/80c37d7d6753485f81497570625d06f0

ragoso:
Hi guys, this fix is interesting... how we can progress with this?

PR #615, opened 2021-12-26
fix: optimization scrollLeft and scrollTop assignment performance

Janlaywss:
When I re-rendered the table, I found that Recalculate Style was very time-consuming。
image

I guess the general reason is: when I slide the table, the columns of the table will be regenerated and rendered, and the sliding of the virtual hidden column in the middle also needs to follow the rhythm.

And I guess that the layout tree of the browser is laid out from top to bottom, and it will be very time-consuming to calculate the scrollLeft.

The solution is to throw the re-assignment calculation of scrollLeft and scrollTop into requestAnimationFrame. Let the reassignment operation not block the current react rendering process.
image

### DIFF
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -377,10 +377,14 @@ export default function createGridComponent({
               break;
           }
         } else {
-          outerRef.scrollLeft = Math.max(0, scrollLeft);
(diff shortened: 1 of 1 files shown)


Janlaywss:
@bvaughn

PR #631, opened 2022-03-21
Fix ARIA violation: add innerElementRole property

jussirantala:
innerElementRole property added for the user to be able to set role. Without this it is not possible to pass ARIA violation checks. For example you could set innerElementRole to "rowgroup" and role for each list item to "row"  and the screen reader would understand that this is list.

### DIFF
--- a/src/createListComponent.js
+++ b/src/createListComponent.js
@@ -1,11 +1,11 @@
 // @flow
 
-import memoizeOne from 'memoize-one';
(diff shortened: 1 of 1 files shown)


Terence-Cheng:
I also had this problem, but I solve it with `innerElementType` like below.

```
        <List
            innerElementType={(innerProps) => (
                <div
                    {...innerProps}
                    role="rowgroup"
                />
            )}

      ....

```

Issue #689, opened 2022-11-30
scrollDirection state is incorrect for RTL lists


DevinFrenze:
I think the `scrollDirection` state is incorrect for fixed and variable sized lists when `direction="rtl"` . Whether I am scrolling right or left, the `scrollDirection` state is "backward". This interfere's with the list's overscan rendering.

Tested on MaxOS Chrome, Safari, and Firefox. I think this applies to any browser for which the "RTLOffsetType" is "negative".

I think I know the reason and will submit a pull request with a potential fix.

PR #690, opened 2022-11-30
fix scrollDirection when direction is RTL

DevinFrenze:
To determine the `scrollDirection` in the `_onScrollHorizontal` method it compares the previous state's `scrollOffset` with the current `scrollLeft` value of the target. For browsers whose "RTLOffsetType" is "negative", this results in the `scrollDirection` being set to "backward".

The `scrollOffset` state value has been converted to a positive number or zero, and the `scrollLeft` value of the target (when "RTLOffsetType" is "negative") is a negative number. When comparing `prevState.scrollOffset < scrollLeft` it is always false because a positive value or zero is never greater than a negative value.

Proposed fix: compare `prevState.scrollOffset < scrollOffset` because `scrollOffset` has already been transformed in the same way that `prevState.scrollOffset` was transformed.

This pull request is intended to address this bug https://github.com/bvaughn/react-window/issues/689.

### DIFF
--- a/src/createListComponent.js
+++ b/src/createListComponent.js
@@ -580,7 +580,7 @@ export default function createListComponent({
         return {
           isScrolling: true,
           scrollDirection:
-            prevState.scrollOffset < scrollLeft ? 'forward' : 'backward',
+            prevState.scrollOffset < scrollOffset ? 'forward' : 'backward',
           scrollOffset,
           scrollUpdateWasRequested: false,
         };


DevinFrenze:
@bvaughn sorry to ping you, but if you have a moment to glance at this and let me know if you think it's a reasonable change i would appreciate it. i'm using react-window in a project and would much rather get this change in and use the main repo than be using a fork indefinitely.

bvaughn (commented):
I think this change looks reasonable. I haven't worked in this code in many months though so I'm not totally confident in my own assessment.

DevinFrenze:
I tested it in a project that I'm developing and it works well there, and it doesn't change behavior when using react-window left-to-right. Of course, there may be edge cases I don't know about.

bvaughn:
That's fair. Okay!

bvaughn:
@DevinFrenze Please check 1.8.10 (just published) and confirm the fix?

DevinFrenze:
@bvaughn it appears to be working well ! thanks for merging

bvaughn:
No problem. Thanks for verifying!

Issue #701, opened 2023-02-27
How to attach elements inside the List component


trqdz:
How to attach elements inside the List component, which will be displayed after the last node,thanks

iahu:
same here.

Maybe we can add two `before` and `after` options to render custom elements besides of the list.

PR #709, opened 2023-03-31
WIP: add before/after slot hook for List and Grid components

iahu:
Hi, I could like to add two render hooks, which is `before` and `after`, to been able to render elements before or after the inner element.

the use case I can give is, that render a cursor element that can scroll beside with the inner element.

and there is a [feature request](https://github.com/bvaughn/react-window/issues/701) about this MR.

thanks.


iahu:
close it, as I find that, the `outerElementType` api can be used as a custom render function.

Issue #413, opened 2020-02-04
All items re-mount on list change


ryanmcclure4:
I have an issue that seems identical to [this](https://github.com/bvaughn/react-window/issues/186). Each time one of the items in my list is modified, added, or removed, itemData changes. This results in all previously rendered items being re-mounted (not just re-rendered). My code is essentially identical to the examples in terms of form (data and list item differ, but the project is structured the same).

Conceptually, using memoize and React.memo should only prevent re-renders of currently mounted items. Thus, neither of those should have an effect on whether components are re-mounted. What other variables that are factored into the list's decision of whether to re-mount the items? I'm converting this from a project that used to render all items in a list, in which case the items are _not_ re-mounted. So the only difference in this case is the virtual list. I've been stuck on this for a while and feel like I'm missing something...

pgostovic:
I'd love to see this issue get some love. I've got a FixedSizeList with items that contain controlled text inputs such that the input's onChange affects some state outside of the FixedSizeList component. Every text input key press updates the state and triggers a re-render of the List; nothing out of the ordinary, just typical React controlled inputs. But because of the unmount/remount, the text input loses focus immediately. There's no scrolling going on here. Here's a trivial one-item example:

```ts
const ListTest: FC = () => {
  const [value, setValue] = useState('')

  return (
    <FixedSizeList height={150} itemCount={1} itemSize={35} width={300}>
      {({ style }) => (
        <input
          type="text"
          style={style}
          value={value}
          onChange={event => setValue(event.currentTarget.value)}
        />
      )}
    </FixedSizeList>
  )
}
```

Surely, only re-rendering should happen in this case and not fully unmounting and remounting. Any ideas?

bvaughn:
What you're asking about is slightly different, @pgostovic, and seems like a duplicate of #420.

Don't define your item renderer inline. It will get recreated each time the parent component re-renders which will mess with text selection apparently.

pgostovic:
Oh man that totally works -- I feel kind of stupid now 😄. Thanks @bvaughn!

bvaughn:
No worries. It's a common mistake. So much so that I plan to change the API for v2 (if I ever finish v2) so that the way you defined it above would not cause this problem.

VitaminCtea:
[...]

PR #804, opened 2024-12-24
docs: add info about DOM re-creation

WofWca:
To decrease the probability of the following issue happening:
https://github.com/bvaughn/react-window/issues/413#issuecomment-597876562


WofWca:
ф

PR #819, opened 2025-07-24
add React 19 compatibility layer

lmousom:
React 19 has completely removed class components, including PureComponent. In react 19+ projects getting error: `Cannot read properties of undefined (reading 'PureComponent’)` was occurring because the library was trying to import PureComponent from React, which no longer exists in React 19

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-window",
-  "version": "1.8.11",
+  "version": "1.8.12",
   "description": "React components for efficiently rendering large, scrollable lists and tabular data",
   "author": "Brian Vaughn <brian.david.vaughn@gmail.com> (https://github.com/bvaughn/)",
   "contributors": [
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -1,7 +1,7 @@
 // @flow
 
 import memoizeOne from 'memoize-one';
-import { createElement, PureComponent } from 'react';
+import { createElement, PureComponent } from './react19-compat';
 import { cancelTimeout, requestTimeout } from './timer';
 import { getScrollbarSize, getRTLOffsetType } from './domHelpers';
 
--- a/src/createGridComponent.js
+++ b/src/createGridComponent.js
@@ -1,6 +1,7 @@
 // @flow
 
 import memoizeOne from 'memoize-one';
+import React from 'react';
 import { createElement, PureComponent } from './react19-compat';
 import { cancelTimeout, requestTimeout } from './timer';
 import { getScrollbarSize, getRTLOffsetType } from './domHelpers';
--- a/src/createListComponent.js
+++ b/src/createListComponent.js
@@ -2,7 +2,7 @@
 
 import memoizeOne from 'memoize-one';
 import React from 'react';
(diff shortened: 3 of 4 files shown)


lmousom:
@bvaughn For a complete React 19 solution, the library could eventually be rewritten to use functional components with hooks. However, this compatibility layer provides an immediate solution that works with React 19 while maintaining backward compatibility

bvaughn:
Ah, good catch.

>  For a complete React 19 solution, the library could eventually be rewritten to use functional components with hooks.

Yeah, this would be a pretty big lift. I've considered it from time to time but I just haven't had the energy.

bvaughn:
> React 19 has completely removed class components, including PureComponent. In react 19+ projects getting error: `Cannot read properties of undefined (reading 'PureComponent’)` was occurring because the library was trying to import PureComponent from React, which no longer exists in React 19

Actually I'm not sure I understand what you're saying here. I assumed I was just lagging behind, but the [docs](https://react.dev/reference/react/PureComponent) don't mention anything about `PureComponent` being removed in 19, just that it's a legacy API and isn't recommended, and I was just able to create a new 19.1 project without any problem.

lmousom:
> > React 19 has completely removed class components, including PureComponent. In react 19+ projects getting error: `Cannot read properties of undefined (reading 'PureComponent’)` was occurring because the library was trying to import PureComponent from React, which no longer exists in React 19
> 
> Actually I'm not sure I understand what you're saying here. I assumed I was just lagging behind, but the [docs](https://react.dev/reference/react/PureComponent) don't mention anything about `PureComponent` being removed in 19, just that it's a legacy API and isn't recommended, and I was just able to create a new 19.1 project without any problem.

The project runs fine during development, but fails after building

bvaughn:
That sounds like something to do with your project build configuration, to me. I haven't found any mention of `PureComponent` being removed from "react" (or even deprecated) and the team would have likely made a change that large _very visible_.

lmousom:
> That sounds like something to do with your project build configuration, to me. I haven't found any mention of `PureComponent` being removed from "react" (or even deprecated) and the team would have likely made a change that large _very visible_.

I'll try without vite. I’m not entirely sure, but I’d love to revamp the entire library. You might want to check this out : [React 19 Upgrade Guide – Removed Module Pattern Factories](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-module-pattern-factories)

bvaughn:
By the way, I'm almost done with a rewrite (#824) that you might be interested in. Love to hear your feedback.

bvaughn:
Version 2 is fully React 19 compatible.

lmousom:
@bvaughn Wow, I honestly can’t believe you went ahead and rewrote the whole library 🙌. That’s such a massive effort, and it really shows your dedication to keeping this project alive for the community. I only scratched the surface, but seeing you take it all the way into a proper v2 is inspiring.

Huge respect for the work you’ve put in here—it’s the kind of effort that benefits so many of us building on top of React. Really grateful to have been a tiny part of the journey ❤️

bvaughn:
Hey thank you 😄

PR #824, opened 2025-08-08
Version 2

bvaughn:
### User facing changes
- More ergonomic props API
- Auto-memoizing for row/cell renderer and additional context props
- Automatically sizing for `List` and `Grid` (aka no more `AutoSizer`)
- Native TypeScript support (aka no more `@types/react-window`)
- Smaller bundle size
- All new website/documentation ([preview](https://react-window-git-issues-821-brian-vaughns-projects.vercel.app/list/fixed-row-height))

### Pre-release testing via NPM
``` sh
npm install react-window@alpha
```

### Known issues with latest alpha (`2.0.0-alpha.7`)
- [ ] `Grid` may not account for scrollbars spaces when scrolling in some cases


bvaughn:
v2 alpha release is available for testing:

https://www.npmjs.com/package/react-window/v/2.0.0-alpha.0

jurerotar:
Hi! 
First of all, thank you for the work on this rework! I've been testing it and have a question. Is there a way to implement horizontal lists? Previously, Lists had `layout` prop, which, if set to `horizontal`, set width property instead of height. I can't find a way to replicate this behavior. Have I missed something?

bvaughn:
Hi @jurerotar! Thanks for the feedback.

Horizontal scrolling won't be something I support with the `List` component in version 2. In my experience/awareness, it's just not used commonly enough to warrant the added complexity.

I think you should be able to use a `Grid` with only one row for horizontal scrolling though. Give that a shot? Only thin that I'm aware of that won't work quite right at the moment is right-to-left which I'll add a note to myself about.

jurerotar:
Thanks for getting back to me so fast!

I've had the same idea about single-row `Grid` and can confirm it works fine! Thanks! 

I've got another question, this time about `initialScrollLeft` and `initialScrollTop` props on the `Grid` component. These have been removed as well. Any plans on re-adding these, or would you prefer the users to implement this functionality themselves, should they need it?

bvaughn:
No problem. I'm working on a docs example for it now since it will probably come up again.
Screenshot 2025-08-25 at 6 43 22 PM

> I've got another question, this time about initialScrollLeft and initialScrollTop props on the Grid component. These have been removed as well. Any plans on re-adding these, or would you prefer the users to implement this functionality themselves, should they need it?

Ah, good question. I think I'd prefer folks use the `gridRef` API and `onResize` callbacks to implement these! Should be able to do something like this:
```
onResize((size, prevSize) => {
  if (!didMountRef.current) {
    didMountRef.current = true;

    gridRef.current.scrollToCell({...});
  }
}
```

jurerotar:
Awesome, thank you! 🎉 

Just to let you know, I've started hitting a couple of `Error: Maximum update depth exceeded` after adding `useGridCallbackRef`s. I'm sure it's something I'm doing wrong, and I'll investigate this in the morning. 

I'm linking the (unfinished) migration PR in case you decide to take a look: https://github.com/jurerotar/Pillage-First-Ask-Questions-Later/pull/85/files#diff-8673c44e118179393fafa3d2a592a98c16dfbfe81c15ad1165491687db724423

Thank you again for your time answering my questions, really appreciate it!

bvaughn:
> I've started hitting a couple of Error: Maximum update depth exceeded after adding useGridCallbackRefs.

My guess is that you're passing something unstable to `rowProps` (or `cellProps`)

I wonder if there's a way I can detect and help make that more obvious (at least in DEV mode) 🤔

smoores-dev:
Something goofy is happening with the types for `onResize` on the `List` component!

image

It looks like because it's being intersected with `HTMLAttributes<HTMLDivElement>`, Typescript unioning the argument types for the methods. You may need to do something like `Omit<HTMLAttributes<HTMLDivElement>, 'onResize'>` to avoid that!

bvaughn:
@smoores-dev Yeah, sorry. That is one of the things listed as "known issues" with the current release (in the PR description). Hoping to push out the next alpha with that and a few other things (like RTL support) soon.

smoores-dev:
Ah, whoops, sorry for missing that! All good!

jurerotar:
> My guess is that you're passing something unstable to `rowProps` (or `cellProps`)

Struggling a bit with this. I think it might be an issue with `useGridCallbackRef`, because removing the `gridRef` prop resolves the maximum update depth, regardless of the props passed. Here's my current code, I tried following the docs:

```ts
const [xAxisMapRulerRef, setXAxisMapRulerRef] = useGridCallbackRef(null);
const xAxisMapRulerProps = useRef<MapRulerCellProps>({ layout: 'horizontal' });

<Grid<MapRulerCellProps>
  className="scrollbar-hidden"
  gridRef={setXAxisMapRulerRef}
  columnCount={gridSize}
  columnWidth={tileSize}
  rowCount={1}
  rowHeight={20}
  cellProps={xAxisMapRulerProps.current}
  cellComponent={MapRulerCell}
/>
```

Removing the `gridRef` makes the issue go away. Might it be something with how `useGridCallbackRef` is defined? It's currently just an asserted `useState`. 

Doing it like this;

```ts
  const xAxisMapRulerRef = useRef<GridImperativeAPI | null>(null);

  const setXAxisMapRulerRef = useCallback((node: GridImperativeAPI) => {
    if (node === null) {
      return;
    }
    xAxisMapRulerRef.current = node;
  }, []);
```

works though.

Would it make sense for `useGridCallbackRef` be defined as:

```ts
export const useGridCallbackRef = (): [ref: GridImperativeAPI | null,  setRef: (node: GridImperativeAPI) => void] => {
  const ref = useRef<GridImperativeAPI | null>(null);

  const setRef = useCallback((node: GridImperativeAPI) => {
    if (node === null) {
      return;
    }
    ref.current = node;
  }, []);

  return [
    ref.current,
    setRef,
  ]
};
```

You'd then use it like this:

```ts
const [xAxisMapRulerRef, setXAxisMapRulerRef] = useGridCallbackRef();
```

---

Second, would it make sense for `cellProps` to be optional? They're currently required, but their description says `Additional props to be passed to the cellComponent`, which may imply them being optional. I currently only require `columnIndex` and `rowIndex` in my `cellComponent`, which are added automatically, but I still have to set `cellProps` to satisfy the type restriction.

bvaughn:
@jurerotar The problem is that the callback form of `Grid` ref gets updated whenever the `cellProps` change, because of how memoization is wired up for that component at the moment. That will be fixed in the next alpha release though!

> Second, would it make sense for cellProps to be optional?

As for this, they're currently required for two reasons:
1. In my experience, they're almost _always_ needed (and in the very rare case they aren't you can just pass an empty object, `{}`)
2. I had trouble getting the TypeScript generics right for the component if I made them optional 😅

jurerotar:
Awesome, thank you very much! Looking forward to continuing testing v2 after the next alpha release! 

Thanks again for your work!

bvaughn:
Version react-window@2.0.0-alpha.6 just published. Only known issues have to do with Grid RTL and minor scrollbar size alignment issues (both listed in the PR description). I'll try to publish react-window@2.0.0-alpha.7 tomorrow or soon, but wanted to push out a few other fixes in the meantime.

@jurerotar and @smoores-dev This should fix the memoization cycle you mentioned.

bvaughn:
Just published version 2.0.0-alpha.7 with RTL support for `Grid`; docs updated.

PR #830, opened 2025-08-31
Relax `List` `rowHeight` constraint

bvaughn:
I think it would be possible to support optional `rowHeight` for list items (using `ResizeObserver` to lazily measure items) though there are a few challenges I have yet to work out:
- How should `scrollToRow` work?
- How should `rowProps` invalidation work?

### How should `scrollToRow` work?
The most obvious approach to this would be to set an offset based on estimated sizes, but that can lead to bad UX (scroll jumping) when scrolling backwards (the more off the estimated sizes turn out to be, the bigger the scroll-jumps).

You can see examples of this [here](https://github.com/bvaughn/infinite-list-reflow-examples?tab=readme-ov-file):

https://github.com/user-attachments/assets/7c9abc8d-eb49-4a05-93b0-57d2f6295eb0

An unsatisfying solution for this would be to not support that imperative API method for this props configuration.

### How should `rowProps` invalidation work?
Version 1 of this library required external code to manually invalidate row data (by calling an imperative `resetAfterIndex` method). Version 2 of this library auto-invalidates whenever a value in `rowProps` change. This makes for a problem with lazily-measured content: What should happen if `rowProps` changes after the user has scrolled? (Cached sizes and offsets might not be valid anymore, and we can't quickly re-populate the cached bounds.)

An unsatisfying solution for this would be to require explicit invalidation for this props configuration.


bvaughn:
Closing in favor of #852

PR #836, opened 2025-09-04
Version 2.1.0

### LINKED ISSUE #834: Fix accessibility
Current solution is not accessible. The role is on the wrong element, you can't set it as a list, since list should not have a div as a child.

Maybe make it support `as` property? Then user could set it to be `<ul>` and use `<li>` as child. Just ensure there are no divs in between.

### LINKED ISSUE #837: [V2] Grid `onCellsRendered` function index params are not `visible[Row/Column][Start/Stop]Index`
Hello @bvaughn ,

I found the nest inner grid `onCellsRendered` index params are not correct. I had tested on my playground, and I’ve ruled out any mistakes I might have made. So I would like you to help confirm if there are any issues.

## Playground

[Code Sandbox](https://codesandbox.io/p/sandbox/react-window-playground-qt53mk)

## Snapshots

Image

Image

Image

bvaughn:
Version 2.1 release

## Improve accessibility through better default ARIA attributes and documentation
The primary goal of this commit is to improve ARIA support.

- Add better default ARIA attributes
  - Add `ariaAttributes` prop to row and cell renderers to simplify integration
  - Add recommended best-practices examples to docs
- Remove intermediate `HTMLDivElement` from `List` and `Grid`
  - This may enable more/better custom CSS styling
  - This may also enable adding an optional `children` prop to `List` and `Grid` for e.g. overlays/tooltips

```tsx
// Example of how to use new `ariaAttributes` prop
function RowComponent({
  ariaAttributes,
  index,
  style,
  ...rest
}: RowComponentProps<object>) {
  return (
    <div style={style} {...ariaAttributes}>
      ...
    </div>
  );
}
```

A secondary goal that may shake out of this is the ability to open up a bit more space for customization with two new optional props:
- `tagName`: Defaults to "div" but can be changed to render e.g. `<ul>`
- `children`: Can be used to add custom overlays, tooltips, etc. within the list/grid element

Resolves #834

## Minor changes to `onRowsRendered` and `onCellsRendered` callbacks

These methods will now receive two params– the first for _visible_ rows and the second for _all_ rows (including overscan), e.g.:
```ts
function onRowsRendered(
  visibleRows: {
    startIndex: number;
    stopIndex: number;
  },
  allRows: {
    startIndex: number;
    stopIndex: number;
  }
): void {
  // ...
}

function onCellsRendered(
  visibleCells: {
    columnStartIndex: number;
    columnStopIndex: number;
    rowStartIndex: number;
    rowStopIndex: number;
  },
  allCells: {
    columnStartIndex: number;
    columnStopIndex: number;
    rowStartIndex: number;
    rowStopIndex: number;
  }
): void {
  // ...
}
```

Resolves #837


src/routes/list/examples/ListAriaRoles.example.html
```diff
@@ -0,0 +1,21 @@
...
+    role="listitem"
+    aria-posinset="2"
+    aria-setsize="1000"
+  >
+    Row 2
+  </div>
+
+  <!-- More rows ... -->
+</div>
+>
```
bvaughn, line 21: Extra character

PR #855, opened 2025-09-25
Fix ResizeObserver being used in the List even when height is provided

tnajdek:
The `style` prop isn’t properly forwarded from `List` to `useVirtualizer`, hence it’s `undefined` [here](https://github.com/bvaughn/react-window/blob/0b871d63b8543cbfe46a2d19ad9a7eaf25059070/lib/hooks/useResizeObserver.ts#L24) and the subsequent early return fails. I wrote a quick test and a fix.

### DIFF
--- a/lib/components/list/List.test.tsx
+++ b/lib/components/list/List.test.tsx
@@ -4,6 +4,7 @@ import { beforeEach, describe, expect, test, vi } from "vitest";
 import { EMPTY_OBJECT } from "../../../src/constants";
 import {
   disableForCurrentTest,
+  simulateUnsupportedEnvironmentForTest,
   updateMockResizeObserver
 } from "../../utils/test/mockResizeObserver";
 import { List } from "./List";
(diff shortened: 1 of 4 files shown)


lib/components/list/List.test.tsx
```diff
@@ -685,6 +685,29 @@ describe("List", () => {
...
       render(<Test />);
     });
+
+    test("should not require resizeObserver if height is provided", () => {
+      const originalResizeObserver = window.ResizeObserver;
+      window.ResizeObserver = vi.fn(() => ({
+        observe: vi.fn(),
+        unobserve: vi.fn(),
+        disconnect: vi.fn()
+      }));
```
bvaughn, line 695:
```suggestion
      simulateUnsupportedEnvironmentForTest();
```

bvaughn (commented):
Wow. That's kind of an embarrassing oversight.

Mind doing the same change for `Grid`?

tnajdek:
> Mind doing the same change for Grid?

Sure, updated!

bvaughn:
Setting it to null means that the test would throw if resizeobserver is
even instantiated, which is a stronger guarantee that we don’t use the API
at all.

tnajdek:
Fair point, updated!

bvaughn:
Fix published in react-window@2.1.2

PR #866, opened 2025-10-09
Imperative API: get-cell-bounds

bvaughn:
Exploration for #864


bvaughn:
Initial impression: I don't love this change. I'm not convinced it's going to be used by 99.9% of users, and it requires kind of awkward memoization to avoid an infinite loop with `rowProps`/`cellProps`.

Issue #875, opened 2025-10-24
tsc error with 2.2.1


drodil:
After updating to 2.2.1 from 2.0.2 I get the following tsc error:

```
node_modules/react-window/dist/react-window.d.ts:23:10 - error TS2724: '"react"' has no exported member named 'ToggleEventHandler'. Did you mean 'EventHandler'?

23 import { ToggleEventHandler } from 'react';
            ~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/react-window/dist/react-window.d.ts:23
```
Any clue why this is happening? React version is 18.3.1.

bvaughn:
I don't know why you're seeing this specifically, because I don't see it when I make a test project using `react@18.3.1` (and `@types/react@18.3.1`). It does look like the `ToggleEventHandler` type was added in React 19 (types) but ideally that shouldn't really matter. I think the only reason it's even included in this project at all is from my attempt to allow people to pass additional, arbitrary DOM attributes to `List` and `Grid`. I'm not sure of the best way to support that while also avoiding having those extra attributes inlined into my type definitions file.

bvaughn:
Hey @Andarist I wonder if yo have any ideas here? (Also feel free to ignore this if you're busy.)

bvaughn:
This issue has been fixed in #876 and released in version 2.2.2

Thanks for the assist, @Andarist!

PR #876, opened 2025-10-25
Annotae Grid and List return types explicitly to avoid redundant inlined inferred types

Andarist:
fixes https://github.com/bvaughn/react-window/issues/875


bvaughn:
Wow! This is very kind of you. Thanks, @Andarist!

bvaughn:
Just tested this and it works with React 18.3+ and 19.0+ but it fails with React <= 18.2:
```
src/App.tsx:15:25 - error TS2339: Property 'index' does not exist on type '{}'.

15 function RowComponent({ index, style }: RowComponentProps<object>) {
```
The underlying cause of which is:
```
Module '"react"' has no exported member 'JSX'.ts(2305)
```
Replacing the `JSX` return type with `ReactElement<any, any>` causes a different underlying issue for `CellComponentProps` and `RowComponentProps` that also breaks the build:
```
Type 'ReactNode' is not assignable to type 'ReactElement<any, any> | null'.
```
Though I think that one can be fixed by changing the return type annotation for `cellComponent` and `rowComponent` to also be `ReactElement`.

Andarist:
> Wow! This is very kind of you. Thanks, @Andarist!

Always happy to help out :)

Issue #909, opened 2026-04-25
Explicit React keys for rows


makimkin:
### Summary

The v2 `List` component does not expose a way to define explicit React keys for rows, unlike v1's `FixedSizeList` which supported the `itemKey` prop.

### Motivation

In v1, `FixedSizeList` and `VariableSizeList` both exposed an `itemKey` prop that allowed consumers to return a stable, unique key per row:

```tsx
<FixedSizeList
  itemCount={items.length}
  itemSize={22}
  itemKey={index => items[index].id}
>
  {Row}
</FixedSizeList>
```

Without this, react-window defaults to using the row index as the React key. This causes incorrect reconciliation behavior when the list data is:
- **Sorted** — items shift positions but React reuses stale rendered state
- **Filtered** — indices no longer map to the same items
- **Dynamically updated** — insertions or deletions cause wrong rows to be recycled

### Current Workaround

There is no clean workaround with the v2 API. Setting a `key` prop on the root element inside `rowComponent` is silently ignored by React when the component is managed externally. Falling back to v1's `FixedSizeList` defeats the purpose of upgrading to the new API.

### Proposed API

Add a `rowKey` prop to the `List` component (mirroring v1's `itemKey`), accepting a function that receives the row index and returns a `string | number`:

```tsx
<List
  rowComponent={MyRow}
  rowCount={items.length}
  rowHeight={22}
  rowKey={index => items[index].id}
/>
```

Optionally, `rowProps` data could also be passed as a second argument to `rowKey`, consistent with how v1 passed `itemData` to `itemKey`:

```tsx
rowKey={(index, data) => data.ids[index]}
```

### Additional Context

This was a well-used feature in v1 (see issues #90, #107, #406) and its absence in v2 is a regression for any use case involving dynamic or reordered lists.

bvaughn:
> Without this, react-window defaults to using the row index as the React key. This causes incorrect reconciliation behavior when the list data is:

> This was a well-used feature in v1 (see issues https://github.com/bvaughn/react-window/issues/90, https://github.com/bvaughn/react-window/issues/107, https://github.com/bvaughn/react-window/issues/406) and its absence in v2 is a regression for any use case involving dynamic or reordered lists.

[...]

PR #911, opened 2026-07-13
fix: add `rowKey` for `List`

WofWca:
Partially addresses https://github.com/bvaughn/react-window/issues/909.
Why partially? Because this doesn't add the prop
to the `Grid` component.

This is basically a (partial) cherry-pick of
3ffd43c69c3067b645ff11284b646fc756725b5d.

Note that `pnpm compile` didn't work for me (couldn't connect to 3000).


bvaughn:
Thanks for kicking off this PR. I've added `Grid` support as well as unit tests.

> Note that pnpm compile didn't work for me (couldn't connect to 3000).

The "compile" script requires the dev server to be running (`pnp dev`) because it also crawls the site and compiles the local search index. No worries though, I've updated this stuff :smile:

PR #913, opened 2026-09-03
fix: update Grid when `dir` changes

imshunsuke:
When a mounted `Grid` changed its explicit `dir` prop between `"ltr"` and `"rtl"`, `useIsRtl` kept the value from the initial render. Its layout effect only recalculated direction when no explicit `dir` was provided.

This updates the hook to synchronize explicit direction changes while preserving DOM and ancestor direction detection when the prop is unset. A regression test covers both `ltr` → `rtl` and `rtl` → `ltr`.

Tested with the full unit test suite, ESLint, TypeScript, Prettier, and the library and documentation production builds.


bvaughn (approved):
Thanks! This is a nice improvement.

I guess if we wanted to go further we could use a mutation observer to listen for a change to the element's `dir` attribute, but maybe that's overkill 🙃

bvaughn:
Released in react-window@2.3.1

